PR #89, opened 2020-05-01
Auto-triage tickets to core board

chapulina:
Uses this action: https://github.com/technote-space/create-project-card-action/tree/v1

Tickets are triaged by the @osrf-triage account, see the examples:

* Issue: https://github.com/ignitionrobotics/testing/issues/2
* PR: https://github.com/ignitionrobotics/testing/pull/3

It looks like the action is triggered for all issues as long as it's in the `master` branch.

But for PRs, it needs to be in the PR branch. See how the action didn't run for this PR: https://github.com/ignitionrobotics/testing/pull/4

### DIFF
--- a/.github/workflows/triage.yml
+++ b/.github/workflows/triage.yml
@@ -1,0 +1,19 @@
+on:
+  issues:
+    types: [opened]
+  pull_request:
+    types: [opened]
+name: Ticket opened
+jobs:
+  assign:
+    name: Add ticket to inbox
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add ticket to inbox
+        uses: technote-space/create-project-card-action@v1
+        with:
+          PROJECT: Core development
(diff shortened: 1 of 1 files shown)


j-rivero:
@osrf-jenkins run tests

j-rivero:
> But for PRs, it needs to be in the PR branch. See how the action didn't run for this PR: [ignitionrobotics/testing#4](https://github.com/ignitionrobotics/testing/pull/4)

This means that the code needs to previously in the master branch? If that is the case I would say that is expected. If it does mean that the action needs to be among the code changed by a PR then it seems weird to me.

chapulina:
> This means that the code needs to previously in the master branch?

I believe it needs to be in the branch that's making the PR. So we should add the action to all release branches + master, which are expected to be used as base branches for PRs.

Issue #92, opened 2020-05-04
Incremental/no-change builds super slow


catskul:
`ign-cmake` is always the long pole in the tent when building. 

"package_source step" and  "Generating API documentation with Doxygen" anecdotally appear to be partially to blame and seem to run when nothing else has changed.

Not doing anything special. Have seen it when compiling ign-cmake manually (`mkdir build; cd build; cmake ../; make; make;` ...) and also using `colcon build`

mjcarroll:
@catskul A substantial amount of the `ign-cmake` build time can typically be attributed to the multiple sub-packages that it builds, which each require an additional `cmake` invocation.

If you want to accelerate it, you can disable these tests by setting a corresponding command line flag for cmake `-DBUILD_TESTING=0`

```
cmake ..  0.94s user 0.27s system 100% cpu 1.212 total
make  82.20s user 24.29s system 99% cpu 1:46.62 total
make  34.90s user 7.56s system 99% cpu 42.625 total
```

```
cmake .. -DBUILD_TESTING=0  0.59s user 0.20s system 99% cpu 0.788 total
make  0.02s user 0.01s system 99% cpu 0.024 total
make  0.02s user 0.01s system 99% cpu 0.024 total
```

Additionally, when building with `colcon`, you can add a `colcon.pkg` file to specifically disable building tests for ignition-cmake

```
{
  "cmake-args": [
    "-DBUILD_TESTING=0",
  ]
}
```

```
$ colcon build --packages-select ignition-cmake2
Starting >>> ignition-cmake2
Finished <<< ignition-cmake2 [0.19s]   
```

mjcarroll:
I believe this should address your issue, feel free to reopen if it does not.

mxgrey:
I wonder if perhaps `BUILD_TESTING` should be off by default, since it really only matters for the build farm, and it's easy enough for us to explicitly enable it for the build farm.

chapulina:
I believe we've been following CTest's convention, on by default: https://cmake.org/cmake/help/v3.7/module/CTest.html . This also matches what `ament_cmake` does, for example.

mxgrey:
Sure, but I think the convention is causing more harm than good here. The convention expects C/C++ compiled targets that can be built once and then that's it. The `ignition-cmake` tests on the other hand are testing the build system, and it seems they're not smart enough to know when it actually matters for them to be rebuilt.

[...]

PR #96, opened 2020-05-19
Remove deprecated bitbucket pipelines config

mjcarroll:
Signed-off-by: Michael Carroll <michael@openrobotics.org>

PR #97, opened 2020-05-19
Disable long-running buildsystem tests by default

mjcarroll:
From #92, this disables several tests that take a while to run and aren't correctly detected by CMake's change-detection.  This adds a new option `BUILDSYSTEM_TESTING`, which will re-enable these tests when true, and enables it in pipelines.

Tested locally with: 

```
$ cmake .. -DBUILDSYSTEM_TESTING=1
$ time make
make  81.50s user 23.50s system 99% cpu 1:45.22 total
```

```
$ cmake ..
$ time make
make  0.18s user 0.07s system 99% cpu 0.254 total
```

Depends on ignition-tooling/ubuntu-bionic-ci-action#9

### DIFF
--- a/.github/workflows/ci-bionic.yml
+++ b/.github/workflows/ci-bionic.yml
@@ -11,6 +11,20 @@ jobs:
         uses: actions/checkout@v2
       - name: Bionic CI
         id: ci
-        uses: ignition-tooling/ubuntu-bionic-ci-action@master
+        uses: ignition-tooling/ubuntu-bionic-ci-action@cmake_args
         with:
           apt-dependencies: 'pkg-config'
+  bionic-ci-plus:
+    runs-on: ubuntu-latest
+    name: Ubuntu Bionic CI
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Bionic CI
+        id: ci
+        uses: ignition-tooling/ubuntu-bionic-ci-action@cmake_args
+        with:
+          apt-dependencies: 'pkg-config'
+          cmake-args: '-DBUILDSYSTEM_TESTING=True'
+
+
--- a/.github/workflows/ci-bionic.yml
+++ b/.github/workflows/ci-bionic.yml
@@ -11,20 +11,8 @@ jobs:
         uses: actions/checkout@v2
       - name: Bionic CI
         id: ci
-        uses: ignition-tooling/ubuntu-bionic-ci-action@cmake_args
-        with:
-          apt-dependencies: 'pkg-config'
-  bionic-ci-plus:
-    runs-on: ubuntu-latest
-    name: Ubuntu Bionic CI
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-      - name: Bionic CI
-        id: ci
-        uses: ignition-tooling/ubuntu-bionic-ci-action@cmake_args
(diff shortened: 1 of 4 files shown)


mjcarroll:
Note to reviewers, the action PR needs to land first, and then I will correct the branches in the workflow file and reduce it to single build with the full build invocation.

.github/workflows/ci-bionic.yml
```diff
@@ -11,6 +11,20 @@ jobs:
         uses: actions/checkout@v2
       - name: Bionic CI
         id: ci
-        uses: ignition-tooling/ubuntu-bionic-ci-action@master
+        uses: ignition-tooling/ubuntu-bionic-ci-action@cmake_args
```
chapulina, line 14:
```suggestion
        uses: ignition-tooling/ubuntu-bionic-ci-action@master
```

mjcarroll:
> Also we need to look into what happened to the Jenkins builds.

This looks like it built on a `xenial` machine.  That doesn't seem correct, given that the minimum cmake version is way past that now.

chapulina:
> This looks like it built on a xenial machine.

Oh I just saw that this PR builds on top of #96 . See my comment on that PR for why we can't remove the pipelines file just yet.

mjcarroll:
> Oh I just saw that this PR builds on top of #96 . See my comment on that PR for why we can't remove the pipelines file just yet.

Rebased and force-pushed.

mjcarroll:
Unless we get the flag turned on in jenkins, I would expect the other CI jobs to fail because there is no test output.

mjcarroll:
> Unless we get the flag turned on in jenkins, I would expect the other CI jobs to fail because there is no test output.

Ah, this will only impact Windows, because we have a test available in UNIX platforms.

scpeters:
for Ubuntu, I think we could set the `BUILDING_EXTRA_CMAKE_PARAMS` variable in https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/docker/ign_cmake-compilation.bash

scpeters:
here's the cmake invocation line on windows: https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/lib/generic-default-devel-windows.bat#L85

mjcarroll:
Since this passes everything but Windows, I'm going to merge.  Windows CI will go green when this lands: https://github.com/ignition-tooling/release-tools/pull/210

scpeters:
> Since this passes everything but Windows, I'm going to merge

I'd prefer to wait for the release-tools PR to land before merging this

mjcarroll:
> I'd prefer to wait for the release-tools PR to land before merging this

In retrospect, I can't merge it because of the branch protection, so you'll get your wish.

scpeters:
windows build with https://github.com/ignition-tooling/release-tools/pull/210 [Build Status](https://build.osrfoundation.org/job/ignition_cmake-ci-pr_any-windows7-amd64/317/) https://build.osrfoundation.org/job/ignition_cmake-ci-pr_any-windows7-amd64/317/

scpeters:
I've cherry-picked these commits onto `ign-cmake2` in `disable_long_tests_2`:

* https://github.com/ignitionrobotics/ign-cmake/tree/disable_long_tests_2

scpeters:
Can you backport to `ign-cmake2`?

PR #101, opened 2020-06-30
make the OGRE plugin path discovery portable

seanyen:
Instead of checking the platform, check the existence of `pkg-config` to accommodate more environments.


chapulina:
@seanyen , mind signing your commit? We're using DCO on Ignition just like ROS 2. Thanks!

seanyen:
@chapulina Thanks for the reminder! Now I signed my commits.

scpeters:

PR #102, opened 2020-07-09
New macros to help with filter google-test in some platforms

j-rivero:
Across the ignition software we have many cases where tests are not prepared to work on Windows, Mac or both. This PR implements a header providing two macros to be used instead of `TEST` macro from google test:

* `IGN_UTILS_TEST_DISABLED_ON_WIN32` creates a google test that will not be run on Windows (although it will be compiled).
* `IGN_UTILS_TEST_ENABLE_ONLY_LINUX` creates a google test that will only be run on Linux (although it will be compiled on all platforms).

Testing: using this PR together with [ign-gui](https://github.com/ignitionrobotics/ign-gui/compare/chapuilna/failing_tests...chapuilna/failing_tests_ign_cmake?expand=1) in: [Build Status](https://build.osrfoundation.org/view/All/job/ign_gui-pr-win/328/) 
(needs more test filtering but already disabled most of the failures).


include/ignition/utilities/detail/ExtraTestMacros.hh
```diff
@@ -0,0 +1,50 @@
...
+#define IGNITION_UTILITIES_DETAIL_EXTRATESTMACROS_HH
+
+#include <ignition/utilities/SuppressWarning.hh>
+
+#define DETAIL_IGN_UTILS_ADD_DISABLED_PREFIX(x) DISABLED_##x
+
+#if defined _WIN32
+
+  #define DETAIL_IGN_UTILS_TEST_DISABLED_ON_WIN32(TestGroup, TestName) \
+      TEST(TestGroup, DETAIL_IGN_UTILS_ADD_DISABLED_PREFIX(TestName))
```
azeey, line 28: It would be nice to support the `TEST_F` and `TEST_P` varieties of gtest as well. One solution would we to limit the macro to just the test name. So the usage would be 

```c++
TEST(MyTestGroup, IGN_UTILS_TEST_ENABLE_ONLY_LINUX(MyTestName))
TEST_F(MyTestGroup, IGN_UTILS_TEST_ENABLE_ONLY_LINUX(MyTestName))
```

include/ignition/utilities/detail/ExtraTestMacros.hh
```diff
@@ -0,0 +1,50 @@
...
+
+#else
+
+  #define DETAIL_IGN_UTILS_TEST_DISABLED_ON_WIN32(TestGroup, TestName) \
+      TEST(TestGroup, TestName)
+
+#endif
+
+#if defined __linux__
```
azeey, line 37: It would also be nice to have a condition to disable tests only on macOS.

chapulina:
The ABI checker failed, I think `ign-cmake` may need some special care since it doesn't have "C++ code"

include/ignition/utilities/ExtraTestMacros.hh
```diff
@@ -0,0 +1,47 @@
...
+/// https://chromium.googlesource.com/external/github.com/google/googletest/+/HEAD/googletest/docs/advanced.md
+#define IGN_UTILS_TEST_DISABLED_ON_MAC(TestName) \
+  DETAIL_IGN_UTILS_TEST_DISABLED_ON_MAC(TestName)
+
+/// \brief Restrict the execution of the test to just the Linux platform
+/// Other platforms will get the test compiled but it won't be run
+/// as part of the test suite execution.
+/// The macro uses the Disabled_ prefix provided by googletest. See
+/// https://chromium.googlesource.com/external/github.com/google/googletest/+/HEAD/googletest/docs/advanced.md
+#define IGN_UTILS_TEST_ENABLE_ONLY_LINUX(TestName) \
```
azeey, line 44: nit: to keep it consistent with the others, would it make sense to name this `IGN_UTILS_TEST_ENABLED_ONLY_ON_LINUX`?

j-rivero:
This run on Windows check these changes together with ign-gui support: [Build Status](https://build.osrfoundation.org/view/All/job/ign_gui-pr-win/386/) Seems to do what it is expected to do, no test failures.

PR #104, opened 2020-07-11
FindUUID: Export include path as expected by Ignition Libraries

traversaro:
With this fix, UUID headers are correctly found even if UUID is not installed in a system prefix.

Fix https://github.com/ignitionrobotics/ign-cmake/issues/103 .

### DIFF
--- a/cmake/FindUUID.cmake
+++ b/cmake/FindUUID.cmake
@@ -22,8 +22,36 @@ if (UNIX)
   if(NOT UUID_FOUND)
     include(IgnManualSearch)
     ign_manual_search(UUID
-                      HEADER_NAMES "uuid/uuid.h"
-                      LIBRARY_NAMES "uuid libuuid")
+                      HEADER_NAMES "uuid.h"
+                      LIBRARY_NAMES "uuid libuuid"
+                      PATH_SUFFIXES "uuid")
+  endif()
+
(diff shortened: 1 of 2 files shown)


traversaro:
I just realized that this patch is not ideal. If someone calls multiple times `find_package(UUID)`, `/usr` is also added to the include directories,  and at some point `get_filename_component(include_dir_parent "/" DIRECTORY)` is called. It is probably  better to first check if the include directory ends with `uuid`  before adding its parent directory.

chapulina:

chapulina:
@traversaro , from your last comment, I assume you're still working on this PR, right?

scpeters:

traversaro:
> 
> 
> @traversaro , from your last comment, I assume you're still working on this PR, right?

Yes, I need to do a small modification to it, I hope to be able to add it today or tomorrow (CEST).

traversaro:
> 
> 
> > @traversaro , from your last comment, I assume you're still working on this PR, right?
> 
> Yes, I need to do a small modification to it, I hope to be able to add it today or tomorrow (CEST).

The updated version that does not suffer the problems described in https://github.com/ignitionrobotics/ign-cmake/pull/104#issuecomment-657433720 has been pushed to the PR.

PR #105, opened 2020-07-12
FindFreeImage: find vcpkg's FreeImage on *nix

traversaro:
The `freeimage`  library installed by vcpkg is called `FreeImage` also on Linux and macOS, 
so to correctly find it is necessary to find both libraries called `freeimage` and libraries called
`FreeImage` also on Linux and macOS.


chapulina:

PR #109, opened 2020-08-18
👩‍🌾 Export `ignition-cmakeN_INCLUDE_DIRS`

chapulina:
While trying to use the macros from #102 on `ign-rendering`, that library was not being able to find the include `ignition/utilities/ExtraTestMacros.hh`.

Strangely, `ign-gazebo` and `ign-gui` are able to find it fine, see https://github.com/ignitionrobotics/ign-gazebo/pull/209/ and https://github.com/ignitionrobotics/ign-gui/pull/76/.

After digging through the CMake variables, I suspect `ign-gazebo` and `ign-gui` may be getting the correct include directory from another dependency. My initial guess was `ign-tools`, because that's a top-level include (`IGNITION-TOOLS_INCLUDE_DIRS=/home/chapulina/blueprint_ws/install/include`) as opposed to a library-specific include path.

In any case, this PR exposes the standard `_INCLUDE_DIRS` variable for `ign-cmake` so it can be used by other projects.


mxgrey:
Could you point me at the `CMakeLists.txt` for `ign-rendering` that links to the `ignition-cmake#::utilities` target? This sounds like an issue that should be solved in the downstream cmake project, not in this project.

scpeters:
> I suspect `ign-gazebo` and `ign-gui` may be getting the correct include directory from another dependency.

> Could you point me at the CMakeLists.txt for ign-rendering that links to the ignition-cmake#::utilities target? This sounds like an issue that should be solved in the downstream cmake project, not in this project.

I think `ign-gazebo` and `ign-gui` are getting the `ignition-cmake#::utilities` target from ignition-plugin, which links to it publicly in the core and loader components:

* https://github.com/ignitionrobotics/ign-plugin/blob/master/loader/CMakeLists.txt#L12
* https://github.com/ignitionrobotics/ign-plugin/blob/master/core/CMakeLists.txt#L12

So I agree with @mxgrey that the problem is likely that `ign-rendering` is not properly linking to `ignition-cmake#::utilities` when it should be, and we shouldn't need to merge this.

chapulina:
Ah thank you both. I was just using `include_directories` on `ign-rendering` with the new variable:

https://github.com/ignitionrobotics/ign-rendering/compare/ign-rendering2...chapulina/disable_failing

Nice detective work, @scpeters. I searched for an example throughout the libraries and forgot to check `ign-plugin.` That's what I need!

PR #117, opened 2020-10-09
Correct CMake Logic and update cpplint to Python3

### LINKED ISSUE #116: codecheck target uses python instead of python3
It looks like [IgnCodeCheck](https://github.com/ignitionrobotics/ign-cmake/blob/ignition-cmake2_2.5.0/cmake/IgnCodeCheck.cmake#L5) is using `python` instead of `python3`, which may be a problem on Ubuntu 20.04. We should probably switch this to `python3`.

mjcarroll:
For Python3 compatibility, this updates the version of cpplint to the latest version available on their repo: https://github.com/cpplint/cpplint/blob/0f2319741f3407d8638cdc7c41e4fc9bad217f68/cpplint.py

I also adjusted the CMake logic to prefer the Python3 interpreter when available.  This logic can be further improved once CMake 3.12 is available across our supported platforms: https://cmake.org/cmake/help/v3.12/module/FindPython.html

Resolves #116

### DIFF
--- a/cmake/IgnPython.cmake
+++ b/cmake/IgnPython.cmake
@@ -1,4 +1,4 @@
-# Copyright 2014 Open Source Robotics Foundation, Inc.
+# Copyright 2020 Open Source Robotics Foundation, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/cmake/IgnPython.cmake
+++ b/cmake/IgnPython.cmake
@@ -1,0 +1,25 @@
+# Copyright 2014 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Copied from ament/ament_cmake: ament_cmake/ament_cmake_core/cmake/core/python.cmake
+
+set(PYTHON_VERSION "" CACHE STRING
+  "Specify specific Python version to use ('major.minor' or 'major')")
+
+# if not specified otherwise use Python 3
+if(NOT PYTHON_VERSION)
+  set(PYTHON_VERSION "3")
+endif()
+
+find_package(PythonInterp ${PYTHON_VERSION} REQUIRED)
--- a/cmake/IgnCodeCheck.cmake
+++ b/cmake/IgnCodeCheck.cmake
@@ -32,9 +32,7 @@ function(ign_setup_target_for_codecheck)
 
   message(STATUS "Adding codecheck target")
 
-  # Setup the codecheck target
-  add_custom_target(codecheck
-
+  add_custom_target(cppcheck
     # First cppcheck
     COMMAND ${CPPCHECK_PATH} ${CPPCHECK_BASE} ${CPPCHECK_EXTRA} -I ${CPPCHECK_INCLUDE_DIRS} ${CPPCHECK_RULES} `${CPPCHECK_FIND}`
 
@@ -43,9 +41,11 @@ function(ign_setup_target_for_codecheck)
   )
 
   add_custom_target(cpplint
-
     # cpplint cppcheck
     COMMAND ${PYTHON_EXECUTABLE} ${IGNITION_CMAKE_CODECHECK_DIR}/cpplint.py --extensions=cc,hh --quiet `${CPPCHECK_FIND}`
   )
 
+  add_custom_target(codecheck
+    DEPENDS cpplint cppcheck
+  )
 endfunction()
--- a/cmake/IgnCodeCheck.cmake
+++ b/cmake/IgnCodeCheck.cmake
@@ -1,17 +1,12 @@
 # Setup the codecheck target, which will run cppcheck and cppplint.
 function(ign_setup_target_for_codecheck)
+  include(IgnPython)
 
   find_program(CPPCHECK_PATH cppcheck)
-  find_program(PYTHON_PATH python)
   find_program(FIND_PATH find)
   
   if(NOT CPPCHECK_PATH)
     message(STATUS "The program [cppcheck] was not found! Skipping codecheck setup")
-    return()
-  endif()
-  
-  if(NOT PYTHON_PATH)
-    message(STATUS "python not found! Skipping codecheck setup.")
     return()
   endif()
   
@@ -45,9 +40,12 @@ function(ign_setup_target_for_codecheck)
 
     # Second cppcheck
     COMMAND ${CPPCHECK_PATH} ${CPPCHECK_BASE} --enable=missingInclude `${CPPCHECK_FIND}`
+  )
+
+  add_custom_target(cpplint
 
     # cpplint cppcheck
-    COMMAND python ${IGNITION_CMAKE_CODECHECK_DIR}/cpplint.py --extensions=cc,hh --quiet `${CPPCHECK_FIND}`
+    COMMAND ${PYTHON_EXECUTABLE} ${IGNITION_CMAKE_CODECHECK_DIR}/cpplint.py --extensions=cc,hh --quiet `${CPPCHECK_FIND}`
   )
 
 endfunction()
--- a/cmake/IgnUtils.cmake
+++ b/cmake/IgnUtils.cmake
@@ -1676,7 +1676,8 @@ macro(ign_build_tests)
 
     # Find the Python interpreter for running the
     # check_test_ran.py script
-    find_package(PythonInterp QUIET)
+
+    include(IgnPython)
 
     # Build all the tests
     foreach(target_name ${test_list})
--- a/codecheck/cpplint.py
+++ b/codecheck/cpplint.py
@@ -373,7 +373,7 @@ _LEGACY_ERROR_CATEGORIES = [
 # off by default (i.e., categories that must be enabled by the --filter= flags).
 # All entries here should start with a '-' or '+', as in the --filter= flag.
 _DEFAULT_FILTERS = [
-        #'-build/c++11',
+        '-build/c++11',
         '-build/header_guard',
         '-build/include_alpha',
         '-build/include_order',
@@ -6414,6 +6414,10 @@ def FlagCxx11Features(filename, clean_lines, linenum, error):
   if include and include.group(1) in ('cfenv',
                                       'condition_variable',
                                       'fenv.h',
+                                      'future',
+                                      'mutex',
+                                      'thread',
+                                      'chrono',
                                       'ratio',
                                       'regex',
                                       'system_error',
--- a/codecheck/cpplint.py
+++ b/codecheck/cpplint.py
@@ -372,7 +372,22 @@ _LEGACY_ERROR_CATEGORIES = [
 # flag. By default all errors are on, so only add here categories that should be
 # off by default (i.e., categories that must be enabled by the --filter= flags).
 # All entries here should start with a '-' or '+', as in the --filter= flag.
-_DEFAULT_FILTERS = ['-build/include_alpha']
+_DEFAULT_FILTERS = [
+        #'-build/c++11',
+        '-build/header_guard',
+        '-build/include_alpha',
+        '-build/include_order',
+        '-build/include_subdir',
+        '-build/namespaces',
+        '-readability/namespace',
+        '-runtime/indentation_namespace',
+        '-runtime/references',
+        '-whitespace/blank_line',
+        '-whitespace/braces',
+        '-whitespace/indent',
+        '-whitespace/newline',
+        '-whitespace/parens',
+        ]
 
 # The default list of categories suppressed for C (not C++) files.
 _DEFAULT_C_SUPPRESSED_CATEGORIES = [
@@ -6399,10 +6414,6 @@ def FlagCxx11Features(filename, clean_lines, linenum, error):
   if include and include.group(1) in ('cfenv',
                                       'condition_variable',
                                       'fenv.h',
-                                      'future',
-                                      'mutex',
-                                      'thread',
-                                      'chrono',
                                       'ratio',
                                       'regex',
                                       'system_error',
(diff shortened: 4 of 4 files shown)


mjcarroll:
Since we don't call this logic directly in the ign-cmake package, I think it would be worth checking cpplint results on a few of the other packages in the ign suite to make sure that warnings/errors haven't diverged from the version we were using previously.

chapulina (commented):
> checking cpplint results on a few of the other packages

+1

scpeters:
> > checking cpplint results on a few of the other packages
> 
> +1

I see 1967 complaints in `ign-sensors3`, which has no complaints with `sh tools/code_check.sh` or `make codecheck` with the released version of ign-cmake2: https://gist.github.com/scpeters/e3c4a36b9fdb392e7a58f804350792c9

chapulina:
> 1967 complaints

ugh, that seems to complain about our style :grimacing:

scpeters:
> > 1967 complaints
> 
> ugh, that seems to complain about our style

I wonder if we had made changes to cpplint.py?

mjcarroll:
> I wonder if we had made changes to cpplint.py?

I looked at the diff, and it wasn't obvious that we made any changes.  But that's certainly a huge difference.

adlarkin (commented):
This PR seems to fix the issue of cppcheck not working with Ubuntu focal :+1: 

However, similar to what @scpeters [mentioned](https://github.com/ignitionrobotics/ign-cmake/pull/117#issuecomment-706423223), I got 11983 errors when running `make codecheck` on the `main` branch of `ign-gazebo`. If I got back to the released version of `ign-cmake2`, I get no errors when running `make codecheck` on the `main` branch of `ign-gazebo`.

mjcarroll:
@adlarkin and @scpeters I believe this is ready for another look, I updated the filters to more closely match our previous configuration: ebc2fca47d8c2ac7dd6a146a29fef20987f66974

scpeters:
how do you feel about adding a file or readme next to cpplint.py that documents where it came from? it's a little easier to find than looking back in the GitHub PR discussion

mjcarroll:
> how do you feel about adding a file or readme next to cpplint.py that documents where it came from? it's a little easier to find than looking back in the GitHub PR discussion

Ah, I had originally put a permalink to the upstream, but I must have removed it at some point.  I'll add it back in for clarity :+1:

scpeters:
I got the following error when testing `ign-sensors`:

~~~
/data_fast/scpeters/ws/ign-sensors/src/ign-sensors/test/integration/TransportTestTools.hh:20:  <condition_variable> is an unapproved C++11 header.  [build/c++11] [5]
~~~

According to [git blame](https://github.com/cpplint/cpplint/blame/0f2319741f3407d8638cdc7c41e4fc9bad217f68/cpplint.py#L6400), that block was last edited in 2014 (https://github.com/cpplint/cpplint/commit/02af6281c61ddc01302801d29c35297a946998a2). It looks like `mutex`, `thread` and `chrono` are also "unapproved"

mjcarroll:
I already dropped a few of those headers, I'm thinking we can drop that category `build/c++11` altogether?

chapulina:
> I'm thinking we can drop that category build/c++11 altogether?

+1

mjcarroll:
Permalink to upstream version added: https://github.com/ignitionrobotics/ign-cmake/pull/117/commits/32c7f59940c86f8d841802dc06cf879cd99a2b83

scpeters:
given our experience with #112, should we use this as an opportunity to test with more downstream packages?

to test, I think we just need to add a `.github/ci/dependencies.yaml` file in a downstream repo that includes this branch of `ign-cmake`?

scpeters:
> given our experience with #112, should we use this as an opportunity to test with more downstream packages?
> 
> to test, I think we just need to add a `.github/ci/dependencies.yaml` file in a downstream repo that includes this branch of `ign-cmake`?

ok, my logic was backwards. the `dependencies.yaml` contents are compiled first and installed before compiling `ign-cmake`. I thought they would all be added to the same `colcon` workspace, but it seems the "main" package gets compiled separately. thus, we can't use the `dependencies.yaml` to compile dependents

mxgrey:
I think the only way to completely test the modules is to build the whole ignition stack from source on top of the changes from this repo.

mjcarroll:
> ok, my logic was backwards. the dependencies.yaml contents are compiled first and installed before compiling ign-cmake. I thought they would all be added to the same colcon workspace, but it seems the "main" package gets compiled separately. thus, we can't use the dependencies.yaml to compile dependents

@scpeters you said it right, it needs to be tested in a downstream rather than here. I was trying to see if I could coax it to work this way as an experiment.

adlarkin (commented):
I went ahead and built the whole ignition collection (Citadel) from source on bionic and focal. Here's a [gist](https://gist.github.com/adlarkin/5a2e378c1ccb11c94e9254d1651d9efa) showing the result of running `make codecheck` in bionic, and a [gist](https://gist.github.com/adlarkin/673657ea39229f91a6ea128681b1e5f2) for focal.

It looks like none of the errors are style related anymore. Most of the errors are actually good/correct errors that we should fix in my opinion. Probably the most common error is for using transitive include headers (for example, making explicit use of `std::string` in a file but not calling `#include <string>` since another file that was included already includes this header). It'd probably be quite a bit of work to fix the transitive headers, but it would be worth it to avoid potential transitive include dependency breaks, as mentioned in the [google c++ style guide](https://google.github.io/styleguide/cppguide.html#Include_What_You_Use).

scpeters (approved):

scpeters:
thanks for the testing @adlarkin, looks good to me, though we may want to proactively fix these errors before our next ign-cmake2 release

mjcarroll:
Does this mean we can remove `cpplint.py` from all other packages (for example: https://github.com/ignitionrobotics/ign-common/tree/main/tools)

scpeters:
I think we haven't finished migration to `make codecheck` from `sh tools/code_check.sh` in Jenkins due to the following issue:

* https://github.com/ignitionrobotics/ign-cmake/issues/31

adlarkin:
> I went ahead and built the whole ignition collection (Citadel) from source on bionic and focal. Here's a [gist](https://gist.github.com/adlarkin/5a2e378c1ccb11c94e9254d1651d9efa) showing the result of running `make codecheck` in bionic, and a [gist](https://gist.github.com/adlarkin/673657ea39229f91a6ea128681b1e5f2) for focal.
> 
> It looks like none of the errors are style related anymore. Most of the errors are actually good/correct errors that we should fix in my opinion. Probably the most common error is for using transitive include headers (for example, making explicit use of `std::string` in a file but not calling `#include <string>` since another file that was included already includes this header). It'd probably be quite a bit of work to fix the transitive headers, but it would be worth it to avoid potential transitive include dependency breaks, as mentioned in the [google c++ style guide](https://google.github.io/styleguide/cppguide.html#Include_What_You_Use).

I'll go ahead and make connecting PRs in the other ignition libraries that resolve the new codecheck errors. Should I make these fixes for ignition blueprint, and then forward-port these changes after merging so that codecheck passes on all ignition library versions?

chapulina:
> Should I make these fixes for ignition blueprint, and then forward-port these changes after merging so that codecheck passes on all ignition library versions?

Yes, please. Thanks!

adlarkin:
Here are the connecting blueprint PRs that should resolve the codecheck issues introduced by this PR:
* https://github.com/ignitionrobotics/ign-fuel-tools/pull/129
* https://github.com/ignitionrobotics/ign-gazebo/pull/443
* https://github.com/ignitionrobotics/ign-gui/pull/144
* https://github.com/ignitionrobotics/ign-launch/pull/70
* https://github.com/ignitionrobotics/ign-msgs/pull/96
* https://github.com/ignitionrobotics/ign-physics/pull/154
* https://github.com/ignitionrobotics/ign-transport/pull/194
* https://github.com/ignitionrobotics/ign-sensors/pull/57

There are a few repositories that I couldn't fix:
* ign-math:
```
root@097debe81253:~/ws/build/ignition-math6# make codecheck
/root/ws/src/ign-math/src/Kmeans.cc:152:  Empty loop bodies should use {} or continue  [whitespace/empty_loop_body] [5]
Total errors found: 1
CMakeFiles/cpplint.dir/build.make:57: recipe for target 'CMakeFiles/cpplint' failed
make[3]: *** [CMakeFiles/cpplint] Error 1
CMakeFiles/Makefile2:131: recipe for target 'CMakeFiles/cpplint.dir/all' failed
make[2]: *** [CMakeFiles/cpplint.dir/all] Error 2
CMakeFiles/Makefile2:171: recipe for target 'CMakeFiles/codecheck.dir/rule' failed
make[1]: *** [CMakeFiles/codecheck.dir/rule] Error 2
Makefile:236: recipe for target 'codecheck' failed
make: *** [codecheck] Error 2
```
* ign-rendering:
```
root@097debe81253:~/ws/build/ignition-rendering2# make codecheck
/root/ws/src/ign-rendering/ogre/src/OgreRenderEngine.cc:497:  Empty loop bodies should use {} or continue  [whitespace/empty_loop_body] [5]
/root/ws/src/ign-rendering/ogre2/src/Ogre2RenderEngine.cc:480:  Empty loop bodies should use {} or continue  [whitespace/empty_loop_body] [5]
Total errors found: 2
CMakeFiles/cpplint.dir/build.make:57: recipe for target 'CMakeFiles/cpplint' failed
make[3]: *** [CMakeFiles/cpplint] Error 1
CMakeFiles/Makefile2:196: recipe for target 'CMakeFiles/cpplint.dir/all' failed
make[2]: *** [CMakeFiles/cpplint.dir/all] Error 2
CMakeFiles/Makefile2:75: recipe for target 'CMakeFiles/codecheck.dir/rule' failed
make[1]: *** [CMakeFiles/codecheck.dir/rule] Error 2
Makefile:197: recipe for target 'codecheck' failed
make: *** [codecheck] Error 2
```

The codecheck errors produced by these repositories are generated by `do ... while` loops (here's the example in [ign-math](https://github.com/ignitionrobotics/ign-math/blob/230ee79c4611b551d7b0d1c02e69e2de068cf8bd/src/Kmeans.cc#L120-L150)). It looks like the linter sees `while (...);`, and interprets this as an empty `while` loop when it's really part of a `do ... while` loop. @mjcarroll is there a way to modify cpplint to handle this case?

scpeters:
> There are a few repositories that I couldn't fix:

I think we could suppress these with a `// NOLINT` on the same line

adlarkin:
> I think we could suppress these with a `// NOLINT` on the same line

That seemed to do the trick for `ign-math`: https://github.com/ignitionrobotics/ign-math/pull/175

For `ign-rendering`, I'm still seeing the error when using `// NOLINT` - could this be because the `while` extends to multiple lines (example [here](https://github.com/ignitionrobotics/ign-rendering/blob/6b5b02f7745158546bbed98104270f2e9f550298/ogre/src/OgreRenderEngine.cc#L500-L501))?

chapulina:
>  I'm still seeing the error when using // NOLINT - could this be because the while extends to multiple lines 

Did you try using `NOLINT` on both lines?

Otherwise, I think it would be ok to add an empty `{}` to that while loop. If it complains about it being empty, you could add a comment in there?

scpeters:
> > I'm still seeing the error when using // NOLINT - could this be because the while extends to multiple lines
> 
> Did you try using `NOLINT` on both lines?
> 
> Otherwise, I think it would be ok to add an empty `{}` to that while loop. If it complains about it being empty, you could add a comment in there?

it's a do - while loop, so we can't add a `{}` to the loop. I suppose we could add empty braces `{}`, but they wouldn't be connected to the loop. I would prefer `NOLINT`

adlarkin:
> Did you try using `NOLINT` on both lines?

Yeah, but I was still getting the false positive.
 
> Otherwise, I think it would be ok to add an empty `{}` to that while loop. If it complains about it being empty, you could add a comment in there?

Like @scpeters said, I can't add `{}` to the loop since it's a do-while loop. I tried adding empty braces after the loop (see below for an example), but this still didn't solve the issue.
```
    while (renderSys &&
           renderSys->getName().compare("OpenGL 3+ Rendering Subsystem") != 0); 
    {}
```

chapulina:
> it's a do - while loop,

ha, I should have scrolled, I was wondering how that loop ever ended :grimacing: 

If the problem is that the while is split in 2 lines, maybe putting it all in one line is the solution? The `NOLINT` may prevent the complaint about the line being too long...

adlarkin:
> If the problem is that the while is split in 2 lines, maybe putting it all in one line is the solution? The `NOLINT` may prevent the complaint about the line being too long...

That worked! I'll make the PR for `ign-rendering` so that you guys can see how this fix looks before we merge it in :+1:

mjcarroll:
https://github.com/ignitionrobotics/ign-gazebo/pull/443 is the only thing left holding this one up.

adlarkin:
> [ignitionrobotics/ign-gazebo#443](https://github.com/ignitionrobotics/ign-gazebo/pull/443) is the only thing left holding this one up.

All connecting PRs have now been merged, so I believe this PR is ready to merge now.

mjcarroll:
> All connecting PRs have now been merged, so I believe this PR is ready to merge now.

Thanks for doing the work on those!

chapulina:
> All connecting PRs have now been merged

They haven't all been forward-ported though :wink:

adlarkin:
> > All connecting PRs have now been merged
> 
> They haven't all been forward-ported though

Good point! Should I make some PRs that forward-port them, or do any of the other branches need to wait before forward porting?

chapulina:
> Should I make some PRs that forward-port them

That would be very helpful! I'm forward-porting `ign-gazebo` now, but I think all other libraries should be ready to go. Thanks!

cmake/IgnPython.cmake
```diff
@@ -0,0 +1,25 @@
...
+
+set(PYTHON_VERSION "" CACHE STRING
+  "Specify specific Python version to use ('major.minor' or 'major')")
+
+# if not specified otherwise use Python 3
+if(NOT PYTHON_VERSION)
+  set(PYTHON_VERSION "3")
+endif()
+
+find_package(PythonInterp ${PYTHON_VERSION} REQUIRED)
```
scpeters, line 25: does this mean `python` is now a required dependency of `ignition-cmake2`? the homebrew bottle build of 2.6.0 in https://github.com/osrf/homebrew-simulation/pull/1222 is failing

[Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_triggered_bottle_builder/label=osx_mojave/124/) https://build.osrfoundation.org/job/generic-release-homebrew_triggered_bottle_builder/label=osx_mojave/124/

scpeters: I think we might want this to be `QUIET` instead of `REQUIRED`

adlarkin: > does this mean python is now a required dependency of ignition-cmake2?

I think that this is what this line implies, although I'm not sure if we meant to create this dependency or not.

> I think we might want this to be QUIET instead of REQUIRED

If we do need python though, what would the implications be of replacing `REQUIRED` with `QUIET`? I'm wondering if marking this `QUIET` can cause issues elsewhere if python is not found. @mjcarroll, what are your thoughts?

mxgrey: If it's an optional dependency that's only needed for optional modules (e.g. testing?), then whatever modules depend on it should check `${PythonInterp_FOUND}` to conditionally build only if this package was found. If the package was found, then the module should skip being built.

scpeters: we can continue the discussion in https://github.com/ignitionrobotics/ign-cmake/pull/132, which proposes to revert python3 to an optional dependency

PR #120, opened 2020-10-22
Revert "Backport #112 to ign-cmake2 (find_package fixes for TINYXML2 and ZeroMQ)"

mjcarroll:
Reverts ignitionrobotics/ign-cmake#115

This broke ign-transport starting: https://build.osrfoundation.org/job/ignition_transport-ci-ign-transport8-windows7-amd64/23/

### DIFF
--- a/cmake/FindTINYXML2.cmake
+++ b/cmake/FindTINYXML2.cmake
@@ -14,21 +14,11 @@
 # limitations under the License.
 #
 ########################################
-# Find tinyxml2.
+# Find tinyxml2. Only debian distributions package tinyxml with a pkg-config.
 
-# Try to find it in Config mode first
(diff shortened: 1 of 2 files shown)


chapulina (approved):
It should be mentioned that this also broke `ign-fuel-tools`, which uses TinyXML2:

https://build.osrfoundation.org/job/ignition_fuel-tools-ci-ign-fuel-tools4-windows7-amd64/

This fix works with https://github.com/ignitionrobotics/ign-transport/pull/191.

@joxoby , heads up that this is being reverted to fix our CI. Please advise in case you have a new solution that satisfies both your use case and our CI.

chapulina:
I think we should also revert #112, although that's less urgent.

PR #123, opened 2020-11-02
Add ImplPtr header

mjcarroll:
This is taken from the [rmw_utils implementation](https://github.com/osrf/rmf_core/blob/master/rmf_utils/include/rmf_utils/impl_ptr.hpp) based on this blog post: http://oliora.github.io/2015/12/29/pimpl-and-rule-of-zero.html

This provides `ignition::utilities::unique_impl_ptr` and `ignition::utilities::impl_ptr`, which can be used anywhere that we are currently using PIMPL, depending on if the class can be copied or not.  This would help clean up several places across the stack where we have non-existent move operators or rule-of-five problems.

Some additional context from @mxgrey

> Using a custom smart pointer to facilitate the PIMPL pattern would be really useful here.
>
> The problem is that our current PIMPL pointers have nothing against providing a mutable reference to the thing they point to, because the built-in const-correctness of C++ only ensures physical constness, with no regard for logical constness.
>
> Using a smart pointer that's specifically designed for the PIMPL pattern prevents this, because its dereference operator is designed to only allow mutable access when the smart pointer is mutable, so when the smart pointer is const it will only provide access to a const-reference.

### DIFF
--- a/include/ignition/utilities/ImplPtr.hh
+++ b/include/ignition/utilities/ImplPtr.hh
@@ -1,0 +1,498 @@
+/*
+ * Copyright (C) 2019 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+#ifndef IGNITION__UTILITIES__IMPLPTR_HH
+#define IGNITION__UTILITIES__IMPLPTR_HH
+
+#include <memory>
+#include <type_traits>
+#include <cassert>
+
+// These utilities were derived from https://github.com/oliora/samples/blob/master/spimpl.h
+// Which uses the following boost license:
+/*
+    Copyright (c) 2015 Andrey Upadyshev (oliora@gmail.com)
+    Distributed under the Boost Software License, Version 1.0.
+    See http://www.boost.org/LICENSE_1_0.txt
+ */
+
+
+namespace ignition {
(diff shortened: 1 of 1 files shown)


mxgrey:
I actually did a lot of heavy lifting on this a while back, which you can see in this [utilities branch](https://github.com/ignitionrobotics/ign-cmake/blob/utilities/include/ignition/utilities/ImplPtr.hh). I even adapted the code to match the ignition code style, since I think someone that I talked to considered that a blocker at the time. I think my move to Singapore probably derailed any effort to adopt this.

I have no problem with using a more direct copy of Andrey Upadyshev's implementation, but in case anyone has a complaint about code style mismatch, you can consider the version that I made back in 2018.

mjcarroll:
> I actually did a lot of heavy lifting on this a while back, which you can see in this utilities branch. I even adapted the code to match the ignition code style, since I think someone that I talked to considered that a blocker at the time. I think my move to Singapore probably derailed any effort to adopt this.

Ah, great, I saw a bitbucket link, and didn't check to see if it had made the transition.  I'm personally good either way on the style front, and just choose to use the one from `rmf_utils` since it looked to be exercised.

chapulina:
+1 to updating to the Ignition style if we're going to release this as part of the Ignition API.

I believe there was also a concern about whether `ign-cmake` is the best place for this. We're starting to put a lot of non-cmake functionality into this library just because it's widely used across the stack, which is less than ideal.

Just checking in case something has changed in the past couple years: do we still need to roll out our own PIMPL solution? Isn't there anything widely available that we could leverage (maybe "vendor" as opposed to copy)?

Finally, if we're going to provide our own implementation, we should have tests for it.

scpeters:
I agree that we should have tests for this, but I'm not opposed in principle to storing this in the utilities folder alongside `SuppressWarning.hh` since this is a header-only implementation

mxgrey:
> Isn't there anything widely available that we could leverage (maybe "vendor" as opposed to copy)?

Since these implementations are usually just a single header file, and C++ packages are a major pain to formally release and distribute, all of the implementations I've seen are made for you to just copy/paste the one header file into your own source code.

I wouldn't rule out the possibility that such a package exists, but I haven't come across one yet.

It's certainly an option for us to vendor it.

mjcarroll:
Based on discussions with the Ignition team, this is likely going to land in ign-common rather than here.

PR #124, opened 2020-11-08
Fix FindIgnOgre on Windows when not using vcpkg

traversaro:
The current version of FindIgnOgre contains a vcpkg-specific workaround on the location of the OgreMain library, that is fixed to point to the "${OGRE_LIBRARY_DIRS}/manual-link/" directory, that it is its location when Ogre is installed by vcpkg. 

However, this creates a problem when installing Ogre with other package managers  on Windows, for example when using conda-forge or conan. To mitigate this problem, we only set the location of the OgreMain library to be "${OGRE_LIBRARY_DIRS}/manual-link/" only if this directory exists, that is only true for vcpkg installations.

### DIFF
--- a/cmake/FindIgnOGRE.cmake
+++ b/cmake/FindIgnOGRE.cmake
@@ -178,7 +178,7 @@ else()
         set(prefix "")
 	# vcpkg uses special directory (lib/manual-link/) to place libraries
 	# with main sysmbol like OgreMain.
-	if(ogre_lib MATCHES "OgreMain" AND NOT IS_ABSOLUTE "${ogre_lib}")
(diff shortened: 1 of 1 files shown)


traversaro:
FYI: 
* @Tobias-Fischer : once merged and released, this should remove the need for the patch https://github.com/conda-forge/libignition-cmake0-feedstock/blob/master/recipe/FindIgnOGRE.cmake.patch from the conda-forge recipe
* @joxoby : I did not tested this, but I suspect that the same problem we have in conda-forge probably could affect also conan

traversaro:
Related comment: https://github.com/conda-forge/libignition-cmake0-feedstock/pull/3#issuecomment-594733549 .

j-rivero (approved):
Thanks Silvio. Seems totally safe to me. Let's wait for the new round of CI before merging.

scpeters (approved):
this looks like a small, safe change to me

PR #132, opened 2020-12-10
Revert python to optional dependency

scpeters:
The homebrew update for ignition-cmake 2.6.0 (osrf/homebrew-simulation#1222) has a failing bottle build:

* [Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_triggered_bottle_builder/label=osx_mojave/124/) https://build.osrfoundation.org/job/generic-release-homebrew_triggered_bottle_builder/label=osx_mojave/124/

It complains about being unable to find Python3. Prior to #117, there was a `find_package(PythonInterp QUIET)` call in `ign_build_tests` when setting up the `check_test_ran.py` test, but that was changed to `REQUIRED`. This reverts that change to restore python to an optional dependency and disables the `cpplint` portion of `make codecheck` if python is not available. We are already checking if python is found before [using the `check_test_ran.py` script](https://github.com/ignitionrobotics/ign-cmake/blob/ign-cmake2/cmake/IgnUtils.cmake#L1699).

The alternative to to make python3 a required dependency, which will require updates to the documentation and packaging metadata.

### DIFF
--- a/cmake/IgnCodeCheck.cmake
+++ b/cmake/IgnCodeCheck.cmake
@@ -40,12 +40,15 @@ function(ign_setup_target_for_codecheck)
     COMMAND ${CPPCHECK_PATH} ${CPPCHECK_BASE} --enable=missingInclude `${CPPCHECK_FIND}`
   )
 
-  add_custom_target(cpplint
-    # cpplint cppcheck
-    COMMAND ${PYTHON_EXECUTABLE} ${IGNITION_CMAKE_CODECHECK_DIR}/cpplint.py --extensions=cc,hh --quiet `${CPPCHECK_FIND}`
+  add_custom_target(codecheck
+    DEPENDS cppcheck
   )
 
-  add_custom_target(codecheck
-    DEPENDS cpplint cppcheck
-  )
+  if(PYTHONINTERP_FOUND)
+    add_custom_target(cpplint
+      COMMAND ${PYTHON_EXECUTABLE} ${IGNITION_CMAKE_CODECHECK_DIR}/cpplint.py --extensions=cc,hh --quiet `${CPPCHECK_FIND}`
+    )
+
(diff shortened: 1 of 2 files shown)


mjcarroll (approved):
Looks good to me, sorry we didn't catch this earlier.

adlarkin (approved):
Looks good to me, thanks for catching this @scpeters! 

I just opened https://github.com/ignition-tooling/action-ignition-ci/pull/26 yesterday so that we can run `make codecheck` in GitHub actions instead of the old `tools/code_check.sh` where applicable, so that we are enforcing the new code checker introduced by #117. This PR shouldn't affect this change to GitHub actions since we only run Ubuntu Bionic on actions, right?

Also, does this mean we'll probably need to do another CMake release to fix homebrew CI? Or is just merging this PR sufficient enough?

chapulina:
> we only run Ubuntu Bionic on actions, right?

We also run Focal from Citadel onwards, and @scpeters started adding macOS actions for some libraries, but that doesn't use `action-ignition-ci`.

> we'll probably need to do another CMake release to fix homebrew CI?

We could make a patch release, I think this fix should benefit Linux as well.

chapulina (approved):

scpeters:
> > we only run Ubuntu Bionic on actions, right?
> 
> We also run Focal from Citadel onwards, and @scpeters started adding macOS actions for some libraries, but that doesn't use `action-ignition-ci`.
> 
> > we'll probably need to do another CMake release to fix homebrew CI?
> 
> We could make a patch release, I think this fix should benefit Linux as well.

yeah I think a patch release would be prudent. I'll open a version bump PR after merging this

scpeters:
> Looks good to me, sorry we didn't catch this earlier.

no worries; I reviewed and approved #117, so I missed it too

PR #138, opened 2020-12-22
Enable make codecheck for ign-cmake

chapulina:
Resolves https://github.com/ignitionrobotics/ign-cmake/issues/134

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -184,7 +184,13 @@ if (BUILD_TESTING AND BUILDSYSTEM_TESTING)
 endif()
 
 # Codecheck
-set(CPPCHECK_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/include)
+set(CPPCHECK_DIRS
+  ${CMAKE_SOURCE_DIR}/examples
+)
+set(CPPCHECK_INCLUDE_DIRS
+  ${CMAKE_SOURCE_DIR}/include
+  ${CMAKE_SOURCE_DIR}/examples
+)
 set(IGNITION_CMAKE_CODECHECK_DIR "${CMAKE_CURRENT_SOURCE_DIR}/codecheck")
 include(IgnCodeCheck)
 ign_setup_target_for_codecheck()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -183,6 +183,12 @@ if (BUILD_TESTING AND BUILDSYSTEM_TESTING)
   add_subdirectory(examples)
 endif()
 
+# Codecheck
+set(CPPCHECK_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/include)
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -183,6 +183,12 @@ if (BUILD_TESTING AND BUILDSYSTEM_TESTING)
   add_subdirectory(examples)
 endif()
 
+# Codecheck
+set(CPPCHECK_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/include)
+set(IGNITION_CMAKE_CODECHECK_DIR "${CMAKE_CURRENT_SOURCE_DIR}/codecheck")
+include(IgnCodeCheck)
+ign_setup_target_for_codecheck()
```
adlarkin, line 190: It looks like GitHub actions isn't happy with this line, but I'm not sure why:
```
  CMake Error at cmake/IgnCodeCheck.cmake:35 (add_custom_target):
  -- Adding codecheck target
    add_custom_target cannot create target "cppcheck" because another target
    with the same name already exists.  The existing target is a custom target
    created in source directory "/github/workspace/examples".  See
    documentation for policy CMP0002 for more details.
    CMakeLists.txt:190 (ign_setup_target_for_codecheck)
```

adlarkin: (The call to `cmake ..` in GitHub actions is currently failing)

mjcarroll: I introduced the `cppcheck` target in a previous PR: https://github.com/ignitionrobotics/ign-cmake/pull/117

This introduced a `cpplint` and `cppcheck` target, which are both executed when you do `make codecheck`

chapulina: Interesting, it worked for me locally. Let me investigate more

chapulina: Ok, I have a fix, pushing soon

chapulina: That `codecheck` target was only being added with `-DBUILDSYSTEM_TESTING=True`. This should fix it, and I also fixed codecheck errors on the examples folder:

9b258c587c46a179e095c7f21dd03e14ce6e0c4f

examples/comp_deps/child/include/empty.hh
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (C) 2018 Open Source Robotics Foundation
```
adlarkin, line 2: This is probably a nit, but should we update the year on these new licenses to 2020? Or does it not really matter?

adlarkin (approved):
Looks good to me, and GitHub actions is happy now! Just one minor question before merging.

examples/comp_deps/child/include/empty.hh
chapulina: I tried to keep the year of when the files were created (note that some files were renamed in 2019).

Also, if you check the `cc` files for those examples, they have copyrights that say 2018. So I think this is ok.

PR #139, opened 2020-12-23
Windows installation via conda-forge

mabelzhang:
Partially addresses https://github.com/ignitionrobotics/docs/issues/117

On Ubuntu, we have the distinction of using `colcon` for top-level compilation, and `cmake` and `make` manually for per-package compilation.
On Windows, those per-package steps up to `cmake` work,but `make` does not readily work. I think we'd need to call Visual Studio's compiler (`cl`?) directly. Instead of that, I currently put in the instructions for `colcon` for per-package as well, since we already know it works. If we figure out some standalone way later, we can change it.

Another reason for using `colcon` for per-package compilation on Windows is that we don't have binary builds for Windows. So users cannot optionally install, say, `ign-*` dependencies from binaries, and only compile the last `ign-` library from source. They would simply have to compile everything from source right now.

The plus side is that colcon will resolve all the dependencies automatically.

The downside is that the colcon workspace means that all the ign-* packages are coupled. So the READMEs for Windows won't look so clean like the Ubuntu sections. For Windows, we will have to refer to dependency ign-* packages and ask the user to install those first. At least that's what I believe; correct me if I'm wrong. Since `ign-cmake` is the lowest dependency, I've included all the setup for Conda and colcon workspace here, and will just refer to ign-cmake from other packages. That means they will not be very self-contained like Ubuntu ones.


traversaro:
> On Windows, those per-package steps up to cmake work,but make does not readily work. I think we'd need to call Visual Studio's compiler (cl?)

CMake provides facilities to compile and install projects without caring for the specific build system used underneath, i.e. in place of `make` you can use: 
~~~
cmake --build . --config Release 
~~~
to compile the project, and instead of `make install` you can use (available only since CMake 3.15, see https://stackoverflow.com/a/56457739):
~~~
cmake --install . --config Release
~~~
The `--config Release` is only useful when you use a Multiple-Config generator (such as Visual Studio or Xcode, see https://cgold.readthedocs.io/en/latest/glossary/multi-config.html) and it will be ignored when using single-config generators such as make or Ninja (see https://cgold.readthedocs.io/en/latest/glossary/single-config.html).

tutorials/install.md
```diff
@@ -85,6 +95,45 @@ Build and install as follows:
     make -j4
     sudo make install
 
+#### Windows
+
+Open a Visual Studio Command Prompt (search for "x64 Native Tools Command Prompt for VS 2019" in the Windows search near the Start button).
```
traversaro, line 100: FYI PR https://github.com/conda-forge/vc-feedstock/pull/19 added the package `vs2019_win-64`, that should source automatically the necessary variables of VS2019, but to be honest I tried it now and it is not working correctly for me, so probably sticking to this well known method is a good idea for now.

traversaro:
> Another reason for using colcon for per-package compilation on Windows is that we don't have binary builds for Windows. So users cannot optionally install, say, ign-* dependencies from binaries, and only compile the last ign- library from source. They would simply have to compile everything from source right now.

Something that you could consider is to install also the ignition dependencies of a given ignition project from conda-forge itself. Some of them already have OR's maintainers, and I think existing maintainers (cc @wolfv @Tobias-Fischer) would be happy to add more mantainers from OR to those packages. Once a new release to a ignition package is done, a PR is automatically generated by the conda-forge infrastructure (see for example https://github.com/conda-forge/libignition-cmake0-feedstock/pull/10), so releasing a new version in conda-forge in practice just means merging that PR, or not even doing that if you tell to the bot that it can auto-merge the release once the test are passing (see https://regro.github.io/cf-scripts/github_actions_infrastructure.html#automerging-prs).

traversaro:
By the way, thanks a lot @mabelzhang for doing this work in supporting Windows also in the docs, it is quite useful!

tutorials/install.md
mabelzhang: "feedstock" is a new word to me. Is the idea that with that package, Visual Studio can be installed via something like `conda install vc vs2019_win-64`, instead of from the Microsoft downloads? And then instead of using the VS Command Prompt, people can use the Anaconda prompt?

mabelzhang:
Cool! Thank you @traversaro for the tips! I tried out the `cmake` commands updated the tutorial accordingly.

I see the conda-forge/libignition-*-feedstock repos. I'll test out them out for subsequent packages (since ign-cmake doesn't have any dependencies).

Tobias-Fischer:
Great to see this happening! As @traversaro mentioned I'm one of the maintainers of some ign packages on conda-forge. We'd be more than happy to add more maintainers and to merge any PRs that improve the feedstocks. Please open relevant PRs and we'll make sure to merge them timely 👍

mabelzhang:
Awesome thanks! conda-forge seems to make things amazingly easy.

tutorials/install.md
traversaro: Sorry for the jargon. In conda, the "feedstock" is the repository that contains the metadata and the scripts to generate a conda package for a given library (see https://conda-forge.org/#about for more info).

The `vs2019_win-64` conda package does not install Visual Studio 2019, but it just automatically invokes the script to set the environment variables set in the `VS Command Prompt` when a conda environment containg it is activated. However, as I was writing in the previous comment is not working correctly (at least for me). In any case if you use the `Visual Studio 2019 16` Generator (as you are doing implicitly now in the docs), instead of the Ninja or NMake generator, it is not necessary to use it as CMake will handle all that configuration automatically.

mabelzhang: Thanks for the explanation! I understood half of that, but sounds good.

traversaro: Feel free to ask for the point in which I was not clear. Re-reading it, I think it would be helpful to point out that the default CMake generator in Windows is the latest Visual Studio one if found, so following the instructions the build will default to use https://cmake.org/cmake/help/v3.19/generator/Visual%20Studio%2016%202019.html .

tutorials/install.md
```diff
@@ -85,6 +124,44 @@ Build and install as follows:
...
+Right-click and run as Administrator.
```
JShep1, line 130: I don't think this needs to be run as admin, everything worked for me without it.

tutorials/install.md
```diff
@@ -85,6 +124,44 @@ Build and install as follows:
...
+1. Navigate to your ``condabin`` if necessary, and then create and activate a Conda environment:
+
+        conda create -n ign-ws
+        conda activate ign-ws
+
+   Once you have activated an environment, a prefix like ``(ign-ws)`` will be prepended to your prompt, and you can use the ``conda`` command outside of ``condabin``.
+
+   You can use ``conda info --envs`` to see all your environments.
+
+   To remove an environment, use``conda env remove --name <env_name>``.
```
JShep1, line 145:
```suggestion
   To remove an environment, use ``conda env remove --name <env_name>``.
```

JShep1 (commented):
Looks good, just a few couple minor tweaks and it should be good.

tutorials/install.md
```diff
@@ -60,6 +89,23 @@ Add OSRF packages:
     wget http://packages.osrfoundation.org/gazebo.key -O - | apt-key add -
     sudo apt-add-repository -s "deb http://packages.osrfoundation.org/gazebo/ubuntu-stable $(lsb_release -c -s) main"
 
+Only on Bionic, update the GCC compiler version:
+
+    sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 800 --slave /usr/bin/g++ g++ /usr/bin/g++-8 --slave /usr/bin/gcov gcov /usr/bin/gcov-8
+
+#### Windows 10
```
JShep1, line 96: I think this Windows section should be deleted, the binary install is above and the source install is below.

tutorials/install.md
```diff
@@ -60,6 +89,23 @@ Add OSRF packages:
...
+
+1. Install [Conda package management system](https://docs.conda.io/projects/conda/en/latest/user-guide/install/download.html).
+   Miniconda suffices.
+
+1. Install [Visual Studio 2019](https://visualstudio.microsoft.com/downloads/).
+   The Community version is free for students, open-source contributors, and individuals.
```
JShep1, line 102: `individuals` sounds a bit awkward here, perhaps personal development?

tutorials/install.md
```diff
@@ -46,6 +46,35 @@ Install Ignition CMake:
 Be sure to replace `<#>` with a number value, such as 1 or 2, depending on
 which version you need.
 
+### Windows
+
+Install [Conda package management system](https://docs.conda.io/projects/conda/en/latest/user-guide/install/download.html).
+Miniconda suffices.
+
+Create if necessary, and activate a Conda environment:
```
JShep1, line 54: Perhaps a note here mentioning the user will need to open a windows command prompt being sure to add `conda` to their PATH?

tutorials/install.md
```diff
@@ -60,6 +89,23 @@ Add OSRF packages:
...
+    sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 800 --slave /usr/bin/g++ g++ /usr/bin/g++-8 --slave /usr/bin/gcov gcov /usr/bin/gcov-8
+
+#### Windows 10
+
+1. Install [Conda package management system](https://docs.conda.io/projects/conda/en/latest/user-guide/install/download.html).
+   Miniconda suffices.
+
+1. Install [Visual Studio 2019](https://visualstudio.microsoft.com/downloads/).
+   The Community version is free for students, open-source contributors, and individuals.
+   Check "Desktop development with C++" in the Workloads tab, and uncheck C++ CMake Tools. We will install cmake via Conda.
```
JShep1, line 103:
```suggestion
   Check "Desktop development with C++" in the `Workloads` tab, and uncheck "C++ CMake Tools." We will install cmake via Conda.
```

tutorials/install.md
JShep1: And also perhaps open-source contributors -> open-source developers?  The first implies to me that a contribution has to have been made to get it for free.

tutorials/install.md
```diff
@@ -85,6 +124,44 @@ Build and install as follows:
     make -j4
     sudo make install
 
+#### Windows
+
+Open a Visual Studio Command Prompt (search for "x64 Native Tools Command Prompt for VS 2019" in the Windows search near the Start button).
```
JShep1, line 129:
```suggestion
Open a Visual Studio Command Prompt (search for "x64 Native Tools Command Prompt for VS 2019" in the Windows search field near the Windows button).
```

tutorials/install.md
```diff
@@ -85,6 +124,44 @@ Build and install as follows:
...
+Right-click and run as Administrator.
+Optionally, right-click and pin to the task bar for quick access in the future.
+
+If you did not add Conda to your ``PATH`` environment variable during Conda installation, you may need to navigate to the location of ``condabin`` in order to use the `conda` command.
```
JShep1, line 133: Is there a particular reason you used double quotes?  Just wondering as most other Ignition docs use single quotes.

tutorials/install.md
mabelzhang: I used the wording verbatim from the Microsoft webpage, which lists Community as "Powerful IDE, free for students, open-source contributors, and individuals." I thought that was the most political correct to just say what they say.

tutorials/install.md
mabelzhang: I remember seeing in some places that the single tick didn't render as verbatim, and having to change to double ticks. I don't remember where that happened, so I just used double ticks to avoid having to fix them, ha. I double-checked the tutorial webpages and the single ticks seem to render correctly. I will change to single ticks.

tutorials/install.md
mabelzhang: Okay. I duplicated the instructions from the Source Install here in 1d50345. It's a little weird to have duplicate instructions but seems that's clearer.

tutorials/install.md
JShep1: Sounds good to me!

JShep1 (approved):
LGTM!

tutorials/install.md
```diff
@@ -85,6 +125,54 @@ Build and install as follows:
...
+1. Install dependencies
+
+        conda install git cmake pkg-config --channel conda-forge 
+
+1. Navigate to where you would like to build the library, and then clone the repository.
+
+        # Optionally, append `-b ign-cmake#` (replace # with a number) to check out a specific version
+        git clone https://github.com/ignitionrobotics/ign-cmake.git
+
+1. Build. Tests do not build at this time, so we turn tests off.
```
chapulina, line 167: Do we know why tests aren't building with this setup? They're building and passing on Windows CI:

https://build.osrfoundation.org/job/ignition_cmake-ci-pr_any-windows7-amd64/405/consoleFull#console-section-7

mabelzhang: I don't know the real reason. I followed the tip from @traversaro here https://github.com/ignitionrobotics/docs/issues/96#issuecomment-723648852
If I don't specify `-DBUILD_TESTING=OFF`, then I get errors like this from a file within the `test` directory:

```
>cmake --build .
Microsoft (R) Build Engine version 16.8.2+25e4d540b for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

  nonvirtual_destructor.cc
Z:\sim\ign_ws\src\ign-cmake\test\warning\nonvirtual_destructor.cc(28,3): error C2220: the follow
ing warning is treated as an error [Z:\sim\ign_ws\build\ignition-cmake2\test\warning\WARNING_non
virtual_destructor.vcxproj]
Z:\sim\ign_ws\src\ign-cmake\test\warning\nonvirtual_destructor.cc(28,3): warning C5205: delete o
f an abstract class 'Base' that has a non-virtual destructor results in undefined behavior [Z:\s
im\ign_ws\build\ignition-cmake2\test\warning\WARNING_nonvirtual_destructor.vcxproj]
Z:\sim\ign_ws\src\ign-cmake\test\warning\nonvirtual_destructor.cc(6): message : see declaration
of 'Base' [Z:\sim\ign_ws\build\ignition-cmake2\test\warning\WARNING_nonvirtual_destructor.vcxpro
j]
Z:\sim\ign_ws\src\ign-cmake\test\warning\nonvirtual_destructor.cc(28,3): message : to simplify m
igration, consider the temporary use of /Wv:18 flag with the version of the compiler with which
you used to build without warnings [Z:\sim\ign_ws\build\ignition-cmake2\test\warning\WARNING_non
Virtual_destructor.vcxproj]
```

mabelzhang:
I noticed that the tutorial has a Documentation and a Testing section... which in other libraries have been living in README instead of the tutorial. Currently, those sections are also present in the README, but are 7 months old. So I think these ones are more recent. Do we want to move these back to the README, or move Documentation and Testing in other Ignition libraries into their installation tutorials for consistency?

tutorials/install.md
JShep1: Hmm, I'm not running into this error and tests are passing for me, not sure why this may be occurring.  I'll probably go ahead and add tests to this tutorial and just have a note indicating that you may need to turn off tests with that flag if you run into issues with them.

mabelzhang: Hmm, I wonder if we have the same CMake versions - would that make a difference?

JShep1: I moved the documentation and testing sections to the readme, it seems a little odd to me that those would reside in the tutorial for installing so I just moved the more information documentation and testing here to the readme d88bce53d207b2ba6b2fcc90bd750950771803d8

JShep1:
> Do we want to move these back to the README, or delete them in README and move Documentation and Testing in other Ignition libraries into their installation tutorials for consistency?

I moved the documentation and testing sections to the README d88bce53d207b2ba6b2fcc90bd750950771803d8

JShep1:
> I moved the documentation and testing sections to the README [d88bce5](https://github.com/ignitionrobotics/ign-cmake/commit/d88bce53d207b2ba6b2fcc90bd750950771803d8)

Reverted in 6da496c24ba7a7ad0a26e3f1cfa53da76fe84fc7 as we're now moving the documentation and testing sections to the install file as it has more context for the new user.

Issue #140, opened 2020-12-29
👨‍🌾 Problem comping on Windows for ign-cmake2


caguero:
https://build.osrfoundation.org/job/ignition_cmake-ci-ign-cmake2-windows7-amd64/57/console

Error message:

```
cl : Command line warning D9025 : overriding '/W3' with '/W4'
nonvirtual_destructor.cc
D:\Jenkins\workspace\ignition_cmake-ci-ign-cmake2-windows7-amd64\ws\ign-cmake\test\warning\nonvirtual_destructor.cc(28): error C2220: the following warning is treated as an error
```

On December 15 the job worked, and since December 22 we have the issue.

chapulina:
I think this is the new warning:

```
D:\Jenkins\workspace\ignition_cmake-ci-ign-cmake2-windows7-amd64\ws\ign-cmake\test\warning\nonvirtual_destructor.cc(28): warning C5205: delete of an abstract class 'Base' that has a non-virtual destructor results in undefined behavior
```

This is probably caused by an upgrade of Visual Studio on the CI machines. That warning was introduced in [VS 2019 v16.5](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warnings-by-compiler-version?view=msvc-160#warnings-introduced-in-visual-studio-2019-version-165-compiler-version-1925286100). That CI machine (` win-windows_local.win8`) is running v16.8.3. While testing a fix, I noticed that `win-windows_nuc.win10` is running v16.4.3, so this may look like a flaky test depending on where it runs.

I have a potential fix in https://github.com/ignitionrobotics/ign-cmake/pull/141.

mjcarroll:
#141 is merged

chapulina:
CI is :large_blue_circle: 

https://build.osrfoundation.org/job/ignition_cmake-ci-ign-cmake2-windows7-amd64/

PR #141, opened 2020-12-29
Suppress warning C5205 on Windows

chapulina:
Attempting a fix for #140.

I queued a couple of test runs, but even though `win-windows_local.win8` was idle (which uses VS2019 v16.8.3), they all waited to run on `win-windows_nuc.win10` (which uses VS2019 v16.4.3), so I couldn't confirm that the fix works. Both CI machines seem to have the same labels, I don't know why the jobs only wanted to run on win10...

CI apart, I wonder if suppressing the new warning is the correct thing to do. It feels like VS2019 is trying to tell us something and we don't want to listen :hear_no_evil:


chapulina:
I just noticed that the job for this PR is also queuing waiting to run on win10, so I think we won't get to confirm the fix until we force it to run on win8.

caguero:
It looks like we have our own internal test that triggers this warning and the expectation is that `IGN_UTILS_WARN_IGNORE__NON_VIRTUAL_DESTRUCTOR` should suppress it. I don't know why we need it but suppress it seems to be the goal here. Maybe @mxgrey can clarify?

mjcarroll:
The only place I believe that this is used (besides the test case) is in `ign-plugin`: https://github.com/ignitionrobotics/ign-plugin/blob/4a87904ac0b70385d8e5bb520e84ab11aa2eaba5/register/include/ignition/plugin/detail/Register.hh#L376-L382

scpeters:
warning C5205 was added in Visual Studio 2019 version 16.5

* https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warnings-by-compiler-version?view=msvc-160#warnings-introduced-in-visual-studio-2019-version-165-compiler-version-1925286100

mjcarroll (approved):

PR #144, opened 2021-01-08
Add an option to disable docs when building

mjcarroll:
Currently we rebuild documentation on every build.  This adds an option to disable this (default is enabled).  This is helpful for development environments with repeated builds.

It can be set via colcon's cmake-args

```
colcon build --cmake-args -DBUILD_DOCS=OFF
```

Or via a `defaults.yaml` file:

```
{
  "build": {
    "merge-install": true,
    "symlink-install": true,
    "cmake-args": [
      "--no-warn-unused-cli",
      "-DCMAKE_CXX_COMPILER_LAUNCHER=ccache",
      "-DCMAKE_EXPORT_COMPILE_COMMANDS=1",
      "-DCMAKE_BUILD_TYPE=RelWithDebInfo",
      "-DBUILD_DOCS=OFF",
    ]
  }
}
```

Signed-off-by: Michael Carroll <michael@openrobotics.org>


mjcarroll:
As an example, I used a simple workspace with `cmake2`, `common3`, and `math6`:

* With Docs:
  * Fresh build: 140 seconds
  * Rebuild: 4.17 seconds
* Without Docs:
  * Fresh build: 139 seconds
  * Rebuild: 0.67 seconds
* Without Docs + CCache (clear cache):
  * Fresh build: 140 seconds
  * Rebuild: 0.67 seconds
* Without Docs + CCache (warm cache):
  * Fresh build: 25.9 seconds
  * Rebuild: 0.67 seconds

cmake/IgnCreateDocs.cmake
```diff
@@ -47,6 +47,12 @@ function(ign_create_docs)
   set(oneValueArgs API_MAINPAGE_MD AUTOGENERATED_DOC TUTORIALS_MAINPAGE_MD)
   set(multiValueArgs "TAGFILES" "ADDITIONAL_INPUT_DIRS")
 
+  option(IGN_BUILD_DOCS "Build docs" ON)
```
azeey, line 50: Any reason we shouldn't make it OFF by default? Also, we have `BUILD_TESTING`, why not just `BUILD_DOCS`?

mjcarroll: > Any reason we shouldn't make it OFF by default?

I was preserving the current behavior.  I would like to see `make docs` a target that could be included/excluded from `all`, but I think that's too much of a shift without a major bump?  Maybe other people feel differently?

> why not just BUILD_DOCS?

Sure, I'm not too attached to this bikeshed color :)

adlarkin: > Any reason we shouldn't make it OFF by default?

I think that if we make it `OFF` by default, then we should probably update the documentation elsewhere to indicate that building documentation is disabled by default so that users don't face issues when following instructions like [these](https://github.com/ignitionrobotics/ign-gazebo#documentation) (in particular, step 3).

> Also, we have BUILD_TESTING, why not just BUILD_DOCS?

Agreed - I think it would make the most sense to either prefix both arguments with `IGN_`, or to take the `IGN_` prefix out of both arguments. One thing to think about is that `BUILD_TESTING` and `BUILD_DOCS` are pretty generic arguments, which could perhaps introduce a naming collision for users that are using Ignition in other projects - so, I think it would be preferable to prefix these arguments with `IGN_`.

mjcarroll: BUILD_TESTING is actually an built-in in CMake as well.

mjcarroll: Updated to be BUILD_DOCS

chapulina (approved):
We should document this somewhere visible. I'd recommend adding a tutorial to `ign-cmake`, but we first need to make sure that's deployed to the docs site (https://github.com/ignitionrobotics/docs/issues/53).

mjcarroll:
> We should document this somewhere visible. I'd recommend adding a tutorial to ign-cmake

Added information in #145

PR #152, opened 2021-02-12
Added port & release to PR templates

jennuine:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,7 +1,12 @@
-<!-- Please remove the appropriate section.
-For example, if this is a new feature, remove the "Bug Fix" section -->
+<!-- 
+Please remove the appropriate section.
+For example, if this is a new feature, remove all sections except for the "New feature" section
 
-# Bug Fix
+If this is your first time opening a PR, be sure to check the contribution guide:
+https://ignitionrobotics.org/docs/all/contributing
+-->
+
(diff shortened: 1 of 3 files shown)


chapulina (approved):
Nice!

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -19,7 +24,7 @@ before and after the PR.-->
 - [ ] All tests passed (See
   [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
```
adlarkin, line 25:
```suggestion
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
```
I think making this text be all on the same line would fix the display issue I mentioned above :crossed_fingers:

adlarkin (commented):
Looks pretty good to me! The one thing I have noticed is that the formatting is a little odd on some of the PRs I have seen using the current templates (a few things in the checklist are scattered across multiple lines). I've attached an image below to show an example (the last three items are scattered across multiple lines). I think that we can fix this by making all of the text for every checklist item be on the same line (I left a comment below that helps better illustrate an example). 

multiple lines template

.github/PULL_REQUEST_TEMPLATE.md
jennuine: Done, thanks! a62736ee5097160bf004f21d785c8ef46e3e4429

PR #154, opened 2021-02-14
glib fix for Windows

chapulina:
# 🦟  Bug fix

## Summary

This is needed to compile `ign-common3` on Windows. [This patch](https://github.com/conda-forge/libignition-cmake0-feedstock/blob/master/recipe/FindGTS.cmake.patch) is applied to the `ignition-cmake2` package on conda-forge, as explained [here](https://github.com/ignitionrobotics/docs/issues/96#issuecomment-742112947). That's why our [Windows source install](https://ignitionrobotics.org/docs/citadel/install_windows_src) tutorial tells users to build everything from source, except for `ign-cmake`. But we should be building everything from source.

I checked that this works on Windows, but I'll leave it as a draft until it can be verified on other platforms too.

## Checklist
- [ ] Added tests


chapulina:
@osrf-jenkins run tests again

chapulina:
Opening for review, I think this shouldn't have adverse effects, because the worst that can happen is that the strings `GLIB_INCLUDE_DIR-NOTFOUND` and `GLIBCONFIG_INCLUDE_DIR-NOTFOUND` get appended to `GTS_INCLUDE_DIRS`. I think that adding invalid include paths shouldn't break anything.

cmake/FindGTS.cmake
```diff
@@ -65,7 +65,14 @@ else()
...
 
   # 2.1 Need glib library
+  find_path(GLIB_INCLUDE_DIR
+    NAMES glib.h
+    PATH_SUFFIXES glib-2.0)
+  find_path(GLIBCONFIG_INCLUDE_DIR
+    NAMES glibconfig.h
+    PATH_SUFFIXES lib/glib-2.0/include)
   find_library(GLIB_LIBRARY glib-2.0)
+  list(APPEND GTS_INCLUDE_DIRS "${GLIB_INCLUDE_DIR}" "${GLIBCONFIG_INCLUDE_DIR}")
```
adlarkin, line 75: > the worst that can happen is that the strings `GLIB_INCLUDE_DIR-NOTFOUND` and `GLIBCONFIG_INCLUDE_DIR-NOTFOUND` get appended to `GTS_INCLUDE_DIRS`. I think that adding invalid include paths shouldn't break anything.

I believe you're correct that invalid include paths shouldn't break anything (they should be "ignored"), but just to be safe, what if we added a check that only adds the include directories if they've been found? We could do something like check to see if `GLIB_INCLUDE_DIR` or `GLIB_CONFIG_INCLUDE_DIR` contain `NOTFOUND` before calling `list(APPEND ...)`.

adlarkin: If believe we could do something like:

```cmake
if (GLIB_INCLUDE_DIR MATCHES "NOTFOUND")
  list(APPEND GTS_INCLUDE_DIRS "${GLIB_INCLUDE_DIR}"
endif()
if (GLIBCONFIG_INCLUDE_DIR MATCHES "NOTFOUND")
  list(APPEND GTS_INCLUDE_DIRS "${GLIBCONFIG_INCLUDE_DIR}"
endif()
```

(https://cmake.org/cmake/help/latest/command/if.html#matches)

chapulina: Good idea, I saw that just doing `if (GLIB_INCLUDE_DIR)` is enough. See 4c9a899

adlarkin: Nice, I didn't realize we could do that - looks good!

cmake/FindGTS.cmake
```diff
@@ -65,7 +65,19 @@ else()
...
+    NAMES glibconfig.h
+    PATH_SUFFIXES lib/glib-2.0/include)
   find_library(GLIB_LIBRARY glib-2.0)
+  if (GLIB_INCLUDE_DIR)
+    list(APPEND GTS_INCLUDE_DIRS "${GLIB_INCLUDE_DIR}")
+  endif()
+  if (GLIBCONFIG_INCLUDE_DIR)
+    list(APPEND GTS_INCLUDE_DIRS "${GLIBCONFIG_INCLUDE_DIR}")
+  endif()
   list(APPEND GTS_LIBRARIES "${GLIB_LIBRARY}")
```
adlarkin, line 81:
```suggestion
  if (GLIB_INCLUDE_DIR)
    list(APPEND GTS_INCLUDE_DIRS "${GLIB_INCLUDE_DIR}")
  endif()
  if (GLIBCONFIG_INCLUDE_DIR)
    list(APPEND GTS_INCLUDE_DIRS "${GLIBCONFIG_INCLUDE_DIR}")
  endif()
  find_library(GLIB_LIBRARY glib-2.0)
  list(APPEND GTS_LIBRARIES "${GLIB_LIBRARY}")
```

Minor comment, but I think it'd be better to append include directories before searching for and including the library. That way, it doesn't appear like the library needs to be searched for first before adding the include directories.

adlarkin: Also, should we do a similar check with the result of the `find_library` call in case that isn't found?

adlarkin: I also wonder if it would be good to display some sort of warning message if the include directories or library can't be found. That way, we can have some debugging information for something like CI.

chapulina: > better to append include directories before searching for and including the library.

I did the other way around, finding the library before appending the includes, because I think the `find_library` call is the one that initially populates `GTS_INCLUDE_DIRS`. See 2d90059.

> should we do a similar check with the result of the find_library call in case that isn't found?

Done on 2d90059

>  I also wonder if it would be good to display some sort of warning message if the include directories or library can't be found.

Honestly I don't know if not finding these directories is a mistake. Our Windows CI has been passing without it. So I think it may be needed by some setups, like with `conda-forge`, but not by others, like with `vcpkg`. If not finding the directories is ok, then a warning could be misleading.

adlarkin (approved):

PR #157, opened 2021-02-24
Find Ogre 2-2

ahcorde:
Signed-off-by: ahcorde <ahcorde@gmail.com>

# 🦟 Bug fix

Related with this issue https://github.com/ignitionrobotics/ign-rendering/issues/223

## Summary

This PR allows to find OGRE 2-2

## Checklist
- [ ] Added tests


mjcarroll (commented):
I think we're going to have to maintain compatibility with both for a little bit, since we aren't bumping the major version on ign-cmake.  This will actually stop OGRE 2.1 from being found.

cmake/FindIgnOGRE2.cmake
```diff
@@ -53,6 +53,14 @@ if (${IgnOGRE2_FIND_VERSION_MAJOR})
   endif()
 endif()
 
+if (${IgnOGRE2_FIND_VERSION_MINOR} VERSION_EQUAL "2")
```
j-rivero, line 56: Thinking on the future and to avoid the assumption that not 2 is 1, could work something like: 
```cmake
message(STATUS "-- Finding OGRE 2.${IgnOGRE2_FIND_VERSION_MINOR}")
set(OGRE2_INSTALL_PATH "OGRE-2.${IgnOGRE2_FIND_VERSION_MINOR}")
```
instead of the whole if block?

mjcarroll: Makes sense

j-rivero (commented):
Testing the thing in a colcon workspace with current ogre-2.1 package + rendering4, the thing compiles just fine. I would not expect regressions with current changes.

We probably need to make some changes to make it work on Windows with different package managers/buildsystems. That could be done in a different PR.

I just notice that the ogre-2.2 example can work with ogre-2.1 using current versions of ign-cmake from .debs returning ogre-2.1 paths. We probably need to implement checks on the consumers of this ign-cmake version (i.e: ign-rendering) to pull these changes into the mix to work correctly.

j-rivero (approved):

PR #163, opened 2021-03-31
Fix hardcoded pkg-config library in examples to fix tests on non-Debian Linux systems

traversaro:
# 🦟 Bug fix

Fixes test failures on platforms (such as non-Debian Linux) in which `CMAKE_INSTALL_LIBDIR` is not equal to `lib`.

## Summary

 ignition-cmake by default installs the pkg-config files in `${CMAKE_INSTALL_LIBDIR}\pkgconfig` (see https://github.com/ignitionrobotics/ign-cmake/blob/b03ca34191e538d334f8d9bb285711a4b487e2a4/cmake/IgnPackaging.cmake#L201). Later, in the tests before this PR it is assumed that they are installed by default in `lib\pkgconfig`. However, there are several platforms (such as a any non-Debian Linux system) in which `${CMAKE_INSTALL_LIBDIR}` is not `lib`, so in those platforms the test would fail.

This PR fixes this by using `${CMAKE_INSTALL_LIBDIR}` also in the tests.

## Checklist
- [x] Added tests

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,4 +1,5 @@
 include(ExternalProject)
+include(GNUInstallDirs)
 
 # test multiple build types
 set(build_types Release RelWithDebInfo Debug)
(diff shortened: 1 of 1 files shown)


traversaro:
Unfortunately the value of `CMAKE_INSTALL_LIBDIR` is not `lib` even on Debian if the install prefix is `/usr`, so a slightly more complex fix is necessary.

examples/CMakeLists.txt
```diff
@@ -174,7 +175,16 @@ if (UNIX)
     "${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME}_fail.xml"
     @ONLY)
   set(_env_vars)
-  list(APPEND _env_vars "PKG_CONFIG_PATH=${example_INSTALL_DIR}/lib/pkgconfig:${FAKE_INSTALL_PREFIX}/lib/pkgconfig:$PKG_CONFIG_PATH")
+  # On Debian and if the install prefix is /usr, the default CMAKE_INSTALL_LIBDIR of this project
+  # as set by GNUInstallDirs will be different from the one of the example project, 
+  # so let's hardcode it in that case
+  if(EXISTS "/etc/debian_version" AND "${CMAKE_INSTALL_PREFIX}" MATCHES "^/usr/?$")
```
adlarkin, line 181: If we want to check that the prefix is `/usr`, should we remove the `?`?

According to the [CMake Regex Specification](https://cmake.org/cmake/help/latest/command/string.html#regex-specification), `?` is defined as `Matches preceding pattern zero or once only`. So, couldn't this mean that there would be zero matches, resulting in the `if` being `true` when it should really be `false`?

traversaro: I guess that the `?` is meant to match both `^/usr$` and `^/usr/$`. Just as a reference, this logic was taken from the relevant code in CMake itself: https://gitlab.kitware.com/cmake/cmake/-/blob/v3.20.0/Modules/GNUInstallDirs.cmake#L242 .

adlarkin:
Changes look okay to me. @j-rivero, mind taking a quick look?

j-rivero (approved):
looks good to me. Thanks @traversaro

PR #168, opened 2021-05-01
Run ign-cmake's copy of check_test_ran

chapulina:
## Summary

Ignition CMake has been running `check_test_ran.py` scripts contained in each Ignition library. But these scripts are all very similar, if not exactly the same. It's much more convenient to maintain a single copy of this script that all libraries can use.

* Update `check_test_ran` to match [the one on ign-gazebo](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo5/tools/check_test_ran.py#L73) so that it prints meaningful test names.
* Install the `tools` directory alongside `benchmark` and `coverage`.
* Note that this also installs `tools/doc_check.sh`, which is not used yet, but once it's installed, it could be used from CMake instead of downloading it from GitHub, see https://github.com/ignition-tooling/action-ignition-ci/pull/30.
* Remove the `tools/code_check.sh` script, which I think doesn't have any use currently.

I don't believe any changes are needed in https://github.com/ignition-release/ign-cmake2-release to install these files, because `benchmark` and `coverage` are already being installed correctly.

## Test it

Install this branch with this PR https://github.com/ignitionrobotics/ign-math/pull/211, which removes that library's own copy of the script.

## Checklist
- [ ] Added tests - no :cry: just manual testing with `ign-math` :point_up: 


scpeters (approved):
This looks good to me. My only thought is a bike-shed about installing to `scripts` instead of `tools`, but this is not important

j-rivero (approved):
Thanks for cleaning up the whole mess. I don't see any problem with the changes in the code.

PR #170, opened 2021-05-06
Port codecov to new configuration

j-rivero:
See https://github.com/ignition-tooling/action-ignition-ci/pull/32/

Signed-off-by: Jose Luis Rivero <jrivero@osrfoundation.org>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
         uses: ignition-tooling/action-ignition-ci@bionic
         with:
           cmake-args: '-DBUILDSYSTEM_TESTING=True'
-          codecov-enabled: true
+          codecov-enabled: false
   focal-ci:
     runs-on: ubuntu-latest
     name: Ubuntu Focal CI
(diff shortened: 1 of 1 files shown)


adlarkin:
Would it be worth merging this and then forward porting it to `main` instead of having a separate PR (#169) for `main`?

adlarkin:
I'm also seeing a failure in GitHub actions:
```
make: *** No rule to make target 'coverage'.  Stop.
```

adlarkin:
@j-rivero friendly ping

j-rivero:
> @j-rivero friendly ping

thanks! I've disabled codecov, if I'm not wrong there is no support in this repository for it

adlarkin (approved):
> I've disabled codecov, if I'm not wrong there is no support in this repository for it

All checks pass, so I'm approving this :+1: @scpeters or @chapulina, beofre we merge this, can one of you confirm that disabling codecov in this repository is okay?

chapulina:
> disabling codecov in this repository is okay?

Yup, I think it should be ok because we aren't generating coverage data either for the tests in `examples` nor the tests in `test`. Not sure if there would be an easy way of adding coverage for CMake files, and I think we shouldn't bother with the utilities C++ code since that's being moved to `ign-utils`.

adlarkin:
Sounds good :+1: merging!

PR #174, opened 2021-07-01
Fix building OGRE / OGRE2 from source in colcon workspace

iche033:
Signed-off-by: Ian Chen <ichen@osrfoundation.org>

# 🦟 Bug fix

## Summary

Problem: When building [ogre2](https://github.com/ignition-forks/ogre-2.1-release) in a colcon workspace, we ran into an error about not finding an ogre **1.x** component. The problem is the cmake module for ogre 1.x is looking into the ogre2 source directory in the colcon workspace instead of the system installed one. 

I also had to fix the values returned by the `OGRE_LIBRARIES` variable. It needs to be the full path to the library. We are actually doing something similar [ogre2 cmake module](https://github.com/ignitionrobotics/ign-cmake/blob/ign-cmake2/cmake/FindIgnOGRE2.cmake#L134) and also [in the logic for ogre 1.x on windows](https://github.com/ignitionrobotics/ign-cmake/blob/ign-cmake2/cmake/FindIgnOGRE.cmake#L166). Not sure why the issue only surfaced in the colcon workspace setup mentioned above.

## Checklist
- [ ] Added tests

### DIFF
--- a/cmake/FindIgnOGRE.cmake
+++ b/cmake/FindIgnOGRE.cmake
@@ -83,6 +83,16 @@ if (NOT WIN32)
 
   # loop through pkg config paths and find an ogre version that is < 2.0.0
   foreach(pkg_path ${PKG_CONFIG_PATH_TMP})
+
+    if (NOT EXISTS ${pkg_path})
+      continue()
+    endif()
+
+    unset(OGRE_FOUND)
+    unset(OGRE_INCLUDE_DIRS)
+    unset(OGRE_LIBRARY_DIRS)
+    unset(OGRE_LIBRARIES)
+
     set(ENV{PKG_CONFIG_PATH} ${pkg_path})
     ign_pkg_check_modules_quiet(OGRE "OGRE >= ${full_version}"
                                 NO_CMAKE_ENVIRONMENT_PATH
@@ -91,6 +101,31 @@ if (NOT WIN32)
       if (NOT ${OGRE_VERSION} VERSION_LESS 2.0.0)
         set (OGRE_FOUND false)
       else ()
+        # set library dirs if the value is empty
+        if (NOT ${OGRE_LIBRARY_DIRS} OR ${OGRE_LIBRARY_DIRS} STREQUAL "")
+          execute_process(COMMAND pkg-config --variable=libdir OGRE
+                          OUTPUT_VARIABLE _pkgconfig_invoke_result
+                          RESULT_VARIABLE _pkgconfig_failed)
+          if(_pkgconfig_failed)
+            IGN_BUILD_WARNING ("Failed to find OGRE's library directory.  The build will succeed, but there will likely be run-time errors.")
+          else()
+            # set ogre library dir and strip line brak
+            set(OGRE_LIBRARY_DIRS ${_pkgconfig_invoke_result})
+            string(REGEX REPLACE "\n$" "" OGRE_LIBRARY_DIRS "${OGRE_LIBRARY_DIRS}")
+
+            string(FIND "${OGRE_LIBRARIES}" "${OGRE_LIBRARY_DIRS}" substr_found)
+            # in some cases the value of OGRE_LIBRARIES is "OgreMain;pthread"
+            # Convert this to full path to library
+            if (substr_found EQUAL -1)
+              foreach(OGRE_LIBRARY_NAME ${OGRE_LIBRARIES})
+                find_library(OGRE_LIBRARY NAMES ${OGRE_LIBRARY_NAME}
+                             HINTS ${OGRE_LIBRARY_DIRS} NO_DEFAULT_PATH)
+                list (APPEND TMP_OGRE_LIBRARIES "${OGRE_LIBRARY}")
+              endforeach()
+              set(OGRE_LIBRARIES ${TMP_OGRE_LIBRARIES})
+            endif()
+          endif()
+        endif()
         break()
       endif()
     endif()
@@ -107,9 +142,10 @@ if (NOT WIN32)
 
     # find ogre components
     foreach(component ${IgnOGRE_FIND_COMPONENTS})
-      ign_pkg_check_modules_quiet(IgnOGRE-${component} "OGRE-${component} >= ${full_version}")
+      ign_pkg_check_modules_quiet(IgnOGRE-${component} "OGRE-${component} >= ${full_version}" NO_CMAKE_ENVIRONMENT_PATH)
       if(IgnOGRE-${component}_FOUND)
         list(APPEND OGRE_LIBRARIES IgnOGRE-${component}::IgnOGRE-${component})
+
       elseif(IgnOGRE_FIND_REQUIRED_${component})
         set(OGRE_FOUND false)
       endif()
--- a/cmake/FindIgnOGRE.cmake
+++ b/cmake/FindIgnOGRE.cmake
@@ -69,7 +69,7 @@ if (NOT WIN32)
                   RESULT_VARIABLE _pkgconfig_failed)
   if(_pkgconfig_failed)
     IGN_BUILD_WARNING ("Failed to get pkg-config search paths")
-  elseif (NOT "_pkgconfig_invoke_result" STREQUAL "")
+  elseif (NOT _pkgconfig_invoke_result STREQUAL "")
     set (PKG_CONFIG_PATH_TMP "${PKG_CONFIG_PATH_TMP}:${_pkgconfig_invoke_result}")
   endif()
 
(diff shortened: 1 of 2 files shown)


iche033:
@darksylinc can you give this a try?

cmake/FindIgnOGRE.cmake
```diff
@@ -83,6 +83,16 @@ if (NOT WIN32)
 
   # loop through pkg config paths and find an ogre version that is < 2.0.0
   foreach(pkg_path ${PKG_CONFIG_PATH_TMP})
+
+    if (NOT EXISTS ${pkg_path})
```
adlarkin, line 87:
```suggestion
  foreach(pkg_path ${PKG_CONFIG_PATH_TMP})
    if (NOT EXISTS ${pkg_path})
```
nit: remove whitespace?

cmake/FindIgnOGRE.cmake
```diff
@@ -91,6 +101,31 @@ if (NOT WIN32)
...
       else ()
+        # set library dirs if the value is empty
+        if (NOT ${OGRE_LIBRARY_DIRS} OR ${OGRE_LIBRARY_DIRS} STREQUAL "")
+          execute_process(COMMAND pkg-config --variable=libdir OGRE
+                          OUTPUT_VARIABLE _pkgconfig_invoke_result
+                          RESULT_VARIABLE _pkgconfig_failed)
+          if(_pkgconfig_failed)
+            IGN_BUILD_WARNING ("Failed to find OGRE's library directory.  The build will succeed, but there will likely be run-time errors.")
+          else()
+            # set ogre library dir and strip line brak
```
adlarkin, line 112:
```suggestion
            # set ogre library dir and strip line break
```

cmake/FindIgnOGRE.cmake
```diff
@@ -107,9 +142,10 @@ if (NOT WIN32)
 
     # find ogre components
     foreach(component ${IgnOGRE_FIND_COMPONENTS})
-      ign_pkg_check_modules_quiet(IgnOGRE-${component} "OGRE-${component} >= ${full_version}")
+      ign_pkg_check_modules_quiet(IgnOGRE-${component} "OGRE-${component} >= ${full_version}" NO_CMAKE_ENVIRONMENT_PATH)
       if(IgnOGRE-${component}_FOUND)
         list(APPEND OGRE_LIBRARIES IgnOGRE-${component}::IgnOGRE-${component})
+
       elseif(IgnOGRE_FIND_REQUIRED_${component})
```
adlarkin, line 149:
```suggestion
      if(IgnOGRE-${component}_FOUND)
        list(APPEND OGRE_LIBRARIES IgnOGRE-${component}::IgnOGRE-${component})
      elseif(IgnOGRE_FIND_REQUIRED_${component})
```
nit: remove whitespace?

cmake/FindIgnOGRE.cmake
```diff
@@ -83,6 +83,16 @@ if (NOT WIN32)
...
   foreach(pkg_path ${PKG_CONFIG_PATH_TMP})
+
+    if (NOT EXISTS ${pkg_path})
+      continue()
+    endif()
+
+    unset(OGRE_FOUND)
+    unset(OGRE_INCLUDE_DIRS)
+    unset(OGRE_LIBRARY_DIRS)
+    unset(OGRE_LIBRARIES)
```
adlarkin, line 94: From the CMake [`unset` documentation](https://cmake.org/cmake/help/latest/command/unset.html):

> Note that when evaluating Variable References of the form `${VAR}`, CMake first searches for a normal variable with that name. If no such normal variable exists, CMake will then search for a cache entry with that name. Because of this unsetting a normal variable can expose a cache variable that was previously hidden. To force a variable reference of the form `${VAR}` to return an empty string, use `set(<variable> "")`, which clears the normal variable but leaves it defined.

I don't believe that the variables here should ever be cache variables, but for the sake of future-proofing the code and also for readability, should we change this to the following?
```suggestion
    set(OGRE_FOUND false)
    set(OGRE_INCLUDE_DIRS "")
    set(OGRE_LIBRARY_DIRS "")
    set(OGRE_LIBRARIES "")
```

adlarkin (commented):
I tested this with `ign-rendering6` (Fortress). When using the `ign-cmake2` branch, I saw the following build warning:

```
CMake Warning at /ogre_source_ws/install/share/cmake/ignition-cmake2/cmake2/IgnConfigureBuild.cmake:53 (message):
   CONFIGURATION WARNINGS:
   -- Skipping component [ogre]: Missing dependency [IgnOGRE] (Components: RTShaderSystem, Terrain, Overlay).
      ^~~~~ Set SKIP_ogre=true in cmake to suppress this warning.
```

When I used the branch in this PR, the warning disappeared, so I believe that this resolves the issue for me.

---

On a related note, I was looking over the `FindIgnOGRE.cmake` file, and believe there may be an issue with the following code block: https://github.com/ignitionrobotics/ign-cmake/blob/f559887e98a2ced7bb89606cb3cfcba4be098ff5/cmake/FindIgnOGRE.cmake#L67-L74

I believe that line 72 should be the following instead (notice how the quotes are removed from `_pkgconfig_invoke_result` so that we can get the value of this variable):
```cmake
elseif (NOT _pkgconfig_invoke_result STREQUAL "")
```

Is this a correct assumption? If so, should we make this fix in this PR as well?

cmake/FindIgnOGRE.cmake
iche033: 3969bcc

cmake/FindIgnOGRE.cmake
iche033: 3969bcc

cmake/FindIgnOGRE.cmake
iche033: yeah I think I noticed that during testing but I opted to follow [what is done](https://github.com/ignitionrobotics/ign-cmake/blob/ign-cmake2/cmake/FindIgnOGRE2.cmake#L278-L280) in FindIgnOGRE2:

I've now changed to setting the variable to empty strings instead of using `unset`, and verified that the variables are indeed empty strings each iteration so I'll stick with your suggestion. I didn't touch the `unset` code in `FindIgnOGRE2` though because don't have a windows machine to test that piece of code.

cmake/FindIgnOGRE.cmake
iche033: 3969bcc

iche033:
> I believe that line 72 should be the following instead (notice how the quotes are removed from _pkgconfig_invoke_result so that we can get the value of this variable):

good catch. I've removed the quotes. 3969bcc

cmake/FindIgnOGRE.cmake
adlarkin: > I didn't touch the unset code in FindIgnOGRE2 though because don't have a windows machine to test that piece of code.

No problem :+1: maybe someone else with easy access to windows can test this. In the meantime, maybe we can at least add a note in `FindIgnOGRE2.cmake` that this could probably be changed to `set(...)` due to the intricacies of how variable references in the form of `${VAR}` work?

iche033: yep added your comment to `FindIgnOGRE2.cmake`. c8a5aea

adlarkin (approved):
I'm approving this since it appears to resolve the build warnings for me, but maybe we can get confirmation from at least one other person before merging.

darksylinc:
I'm afraid it's worse now. Colcon won't complain (ok it does because it doesn't find libOgreHlmsPbs_d and libOgreHlmsUnlit_d.so) but now workarounding the bugs won't fix anything.

Colcon just refuses to install libignition-rendering-ogre2 but says everything is fine.

A couple notes:

I'm building with `colcon build --cmake-args -DBUILD_TESTING=OFF -DCMAKE_BUILD_TYPE=Debug -DBUILD_DOCS=OFF --merge-install` notice CMAKE_BUILD_TYPE=Debug

I'm patching ign-rendering with this so that Debug builds and runs ok:

```
diff --git a/ogre2/src/CMakeLists.txt b/ogre2/src/CMakeLists.txt
index 44b8b5a5..b19a115f 100644
--- a/ogre2/src/CMakeLists.txt
+++ b/ogre2/src/CMakeLists.txt
@@ -1,3 +1,8 @@
+if( UNIX )
+	# lld is much faster than ld for linking
+	set( CMAKE_SHARED_LINKER_FLAGS "-fuse-ld=lld" )
+endif()
+
 # Collect source files into the "sources" variable and unit test files into the
 # "gtest_sources" variable.
 ign_get_libsources_and_unittests(sources gtest_sources)
@@ -33,6 +38,9 @@ target_link_libraries(${ogre2_target}
     ${OPENGL_LIBRARIES}
     IgnOGRE2::IgnOGRE2)
 
+target_compile_definitions(${ogre2_target} PRIVATE $<$<CONFIG:Debug>:DEBUG=1 _DEBUG=1>)
+
+
 set (versioned ${CMAKE_SHARED_LIBRARY_PREFIX}${PROJECT_NAME_LOWER}-${engine_name}${CMAKE_SHARED_LIBRARY_SUFFIX})
 set (unversioned ${CMAKE_SHARED_LIBRARY_PREFIX}${PROJECT_NAME_NO_VERSION_LOWER}-${engine_name}${CMAKE_SHARED_LIBRARY_SUFFIX})
 
```

and only have `libogre-1.9-dev` installed but **not** `libogre-2.1-dev`. I suspect this could be the reason.

Another thing I noticed is that some of the paths were wrong:

Cmake

1. Hlms ones were corrected by me.
2. Overlay is pointing to 1.9's system one
3. pkgcfg is pointing to 1.9's system OgreMain

I will now try a non-debug build to see if that works.

darksylinc:
Btw just add the suggested patch in https://github.com/ignitionrobotics/ign-cmake/pull/174#issuecomment-873486083 and it would be good to go

iche033:
applied patch in ca041f9f640867eb1fa6074d80600137796e52cb, thanks!

adlarkin (approved):
New patch seems good to me, thanks @darksylinc!

PR #175, opened 2021-08-10
Fix problems on IgnOGRE when version is not found

j-rivero:
# 🦟 Bug fix

## Summary

When Ogre is not found (looking for 1.10 in a 1.9 environment), the builds fail with:

```cmake
CMake Error at /usr/share/cmake/ignition-cmake2/cmake2/FindIgnOGRE.cmake:214 (if):
  if given arguments:

    "VERSION_LESS" "1.11.0"

  Unknown arguments specified
  /usr/share/cmake/ignition-cmake2/cmake2/IgnUtils.cmake:189 (find_package)
  CMakeLists.txt:69 (ign_find_package)


-- Configuring incomplete, errors occurred!
```

There is a final block starting at line 240 that is not protected by the check of `OGRE_FOUND` leaving `OGRE_VERSION` empty and making the whole thing to crash. I did not find any reason not to include it under `OGRE_FOUND`.

## Checklist
- [ ] Added tests

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,8 @@
 
 ### Ignition CMake 2.X.X (20XX-XX-XX)
 
+1. Fix problems on IgnOGRE.cmake when version is not found
+
 ### Ignition CMake 2.8.0 (2021-04-30)
 
 1. Fix hardcoded pkg-config library in examples
--- a/cmake/FindIgnOGRE.cmake
+++ b/cmake/FindIgnOGRE.cmake
@@ -236,23 +236,24 @@ else()
     set(OGRE_LIBRARIES ${ogre_all_libs})
     set(OGRE_RESOURCE_PATH ${OGRE_CONFIG_DIR})
   endif()
-endif()
 
-# manually search and append the the RenderSystem/GL path to
-# OGRE_INCLUDE_DIRS so OGRE GL headers can be found
-foreach (dir ${OGRE_INCLUDE_DIRS})
-  get_filename_component(dir_name "${dir}" NAME)
-  if("${dir_name}" STREQUAL "OGRE")
-    if(${OGRE_VERSION} VERSION_LESS 1.11.0)
-      set(dir_include "${dir}/RenderSystems/GL")
+  # manually search and append the the RenderSystem/GL path to
+  # OGRE_INCLUDE_DIRS so OGRE GL headers can be found
+  foreach(dir ${OGRE_INCLUDE_DIRS})
+    get_filename_component(dir_name "${dir}" NAME)
+    if("${dir_name}" STREQUAL "OGRE")
+      if(${OGRE_VERSION} VERSION_LESS 1.11.0)
+        set(dir_include "${dir}/RenderSystems/GL")
+      else()
+        set(dir_include "${dir}/RenderSystems/GL" "${dir}/Paging")
+      endif()
     else()
-      set(dir_include "${dir}/RenderSystems/GL" "${dir}/Paging")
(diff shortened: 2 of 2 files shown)


cmake/FindIgnOGRE.cmake
```diff
@@ -236,23 +236,24 @@ else()
...
-      set(dir_include "${dir}/RenderSystems/GL" "${dir}/Paging")
+      set(dir_include "${dir}")
     endif()
-  else()
-    set(dir_include "${dir}")
-  endif()
-  list(APPEND OGRE_INCLUDE_DIRS ${dir_include})
-endforeach()
+    list(APPEND OGRE_INCLUDE_DIRS ${dir_include})
+  endforeach()
```
iche033, line 254: looks like this block of code is now moved to to the Windows section (instead of `if (OGRE_FOUND)`)? When building ign-rendering with this branch, I get error about not being able to find an ogre include header during compilation. How about just protecting it with a separate `if (OGRE_FOUND)` check?

j-rivero: Ouch, good catch! 0b05939

adlarkin (commented):
I'm trying to test this locally. Any tips for what setup to use to re-produce this behavior?

Changelog.md
```diff
@@ -2,6 +2,8 @@
 
 ### Ignition CMake 2.X.X (20XX-XX-XX)
 
+1. Fix problems on IgnOGRE.cmake when version is not found
```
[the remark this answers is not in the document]
iche033: I'm testing it with this PR: https://github.com/ignitionrobotics/ign-rendering/pull/376

Without these changes, I get the `VERSION_LESS` cmake error

iche033:
@adlarkin , I'm testing it with this PR: https://github.com/ignitionrobotics/ign-rendering/pull/376

Without these changes, I get the `VERSION_LESS` cmake error

adlarkin:
> I'm testing it with this PR: [ignitionrobotics/ign-rendering#376](https://github.com/ignitionrobotics/ign-rendering/pull/376)
> 
> Without these changes, I get the `VERSION_LESS` cmake error

I'm not able to reproduce the error without these changes :thinking: I built a workspace with the `ign-cmake2` branch and https://github.com/ignitionrobotics/ign-rendering/pull/376. are there certain ogre packages I should/should not have installed? I installed all of the dependencies listed in https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering3/.github/ci/packages.apt, which includes `libogre-1.9-dev` and `libogre-2.1-dev`.

iche033:
> I'm not able to reproduce the error without these changes

hmm I just rebuilt my workspace and I can't reproduce it now too.. 

But I think we may run into this error with no ogre debs installed in the system though.

adlarkin:
I think that the changes in the PR make sense. I'd just like to be able to test locally if possible in order to avoid new breaks that we might not be aware of. But, maybe it's difficult to create a scenario to test this, I'm not sure.

iche033 (approved):
I think the changes make sense. We can get this in

Changelog.md
```diff
@@ -1,6 +1,9 @@
 ## Ignition CMake 2.x
 
-### Ignition CMake 2.X.X (20XX-XX-XX)
+### Ignition CMake 2.X.X (2021-08-23)
```
adlarkin, line 3: Are we planning to make an `ign-cmake` release once this PR is merged?

adlarkin (approved):

PR #176, opened 2021-08-12
Special case for ogre2.2 on Windows

mjcarroll:
Fixes finding osrf vcpkg ogre2.2 on Windows.

Signed-off-by: Michael Carroll <michael@openrobotics.org>

### DIFF
--- a/cmake/FindIgnOGRE2.cmake
+++ b/cmake/FindIgnOGRE2.cmake
@@ -293,7 +293,18 @@ else() #WIN32
     set(OGRE_${ogre_component}_FOUND FALSE)
   endforeach()
   # currently designed to work with osrf vcpkg ogre2 portfile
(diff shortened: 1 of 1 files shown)


chapulina:
Is there a CI test run that we can refer to?

mjcarroll:
> Is there a CI test run that we can refer to?

Not really, this also requires the ogre2.2 branch in ign-rendering.  I have tested it locally on my Windows machine.

chapulina (approved):
We're having a hard time setting up custom CI for this, so let's yolo merge and test it with https://github.com/ignitionrobotics/ign-rendering/pull/272. It shouldn't break anyone else, so in the worst case, it doesn't improve anything.

ahcorde:
I will test it tomorrow

PR #178, opened 2021-08-19
Revert "Another try at OGRE2.1/OGRE2.2 on Windows"

mjcarroll:
Reverts ignitionrobotics/ign-cmake#177

This broke CI against OGRE 2.1


chapulina (approved):

chapulina:
Let's get this in to fix downstream CI and work on a fix that satisfies both 2.1 and 2.2 (without breaking 1.X :sweat_smile: )

PR #179, opened 2021-08-19
Fix ogre2 include

mjcarroll:
Alternative to #178, tested on `win-beefy` CI machine against OGRE 2.1.


mjcarroll:
I'm going to lump this with the re-revert

Issue #40, opened 2018-06-15
Target UUID::UUID does not exist


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-cmake/issues/40)) by Anonymous.**

----------------------------------------
I have been trying to update an older ROS-Gazebo project, to gazebo 9 that is using ignition math:

https://github.com/ethz-asl/rotors_simulator 

Currently I am running Ubuntu 18.04, and Gazebo 9.1 and cmake 3.10.2.

When I compile the code get an error from CMAKE, that looks like this:

```
CMake Error at /home/martin/ros_sim_drone/src/rotors_simulator/rotors_gazebo_plugins/CMakeLists.txt:26 (target_link_libraries):
  Error evaluating generator expression:

    $<TARGET_PROPERTY:UUID::UUID,INTERFACE_INCLUDE_DIRECTORIES>

  Target "UUID::UUID" not found.
```

The whole response looks like this:

```
Errors     << rotors_gazebo_plugins:check /home/martin/ros_sim_drone/logs/rotors_gazebo_plugins/build.check.002.log                                                                                        
CMake Warning (dev) at /usr/share/cmake-3.10/Modules/FindBoost.cmake:911 (if):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Quoted variables like "chrono" will no longer be dereferenced when the
  policy is set to NEW.  Since the policy is not set the OLD behavior will be
  used.
Call Stack (most recent call first):
  /usr/share/cmake-3.10/Modules/FindBoost.cmake:1558 (_Boost_MISSING_DEPENDENCIES)
  /usr/share/OGRE/cmake/modules/FindOGRE.cmake:318 (find_package)
  /usr/lib/x86_64-linux-gnu/cmake/gazebo/gazebo-config.cmake:175 (find_package)
  CMakeLists.txt:20 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at CMakeLists.txt:28 (add_library):
  Policy CMP0028 is not set: Double colon in target name means ALIAS or
  IMPORTED target.  Run "cmake --help-policy CMP0028" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.

[...]

PR #193, opened 2021-11-10
FindIgnURDFDOM cmake module

koonpeng:
# 🎉 New feature

## Summary

Adds a minimal `FindURDF.cmake` module, this will help migrate sdformat to ign-cmake ignitionrobotics/sdformat#181.

## Test it

```cmake
cmake_minimum_required(VERSION 3.10.2 FATAL_ERROR)
find_package(ignition-cmake2 REQUIRED)
ign_find_package(URDF REQUIRED)
```

## Checklist
- [ ] Added tests


traversaro:
Could it make sense to call it `FindIgnURDFDOM`? In the past we had problems with collision between ign-cmake provided `Find<pkg>.cmake` and similarly named files distributed by other projects (see for example https://github.com/ignitionrobotics/ign-cmake/issues/40). Naming (at least new) all `Find**.cmake` scripts as `FindIgn***.cmake` should avoid these kind of problems. Furthermore, the library that this scripts finds is called `urdfdom` (see https://github.com/ros/urdfdom and https://repology.org/project/urdfdom/versions), so using urdfdom also in the Find script may avoid confusion.

scpeters:
> Could it make sense to call it `FindIgnURDFDOM`? In the past we had problems with collision between ign-cmake provided `Find<pkg>.cmake` and similarly named files distributed by other projects (see for example #40). Naming (at least new) all `Find**.cmake` scripts as `FindIgn***.cmake` should avoid these kind of problems. Furthermore, the library that this scripts finds is called `urdfdom` (see https://github.com/ros/urdfdom and https://repology.org/project/urdfdom/versions), so using urdfdom also in the Find script may avoid confusion.

I agree; please rename to `FindIgnURDFDOM`

j-rivero (changes requested):
I'm afraid that the current implementation is not valid for Windows/MSVC builds.

Is there any particular reason not to use the urdfdom CMake modules instead of using pkg-config? I think that we could support Windows and other system well by using CMake if possible.

scpeters:
> I'm afraid that the current implementation is not valid for Windows/MSVC builds.
> 
> Is there any particular reason not to use the urdfdom CMake modules instead of using pkg-config? I think that we could support Windows and other system well by using CMake if possible.

the SDFormat search logic has a fall-back to find_package if pkg_check_modules does not find it:

* https://github.com/ignitionrobotics/sdformat/blob/sdformat12_12.1.0/cmake/SearchForStuff.cmake#L13-L51

it looks like DART now uses only `find_package`:

* https://github.com/dartsim/dart/blob/main/cmake/DARTFindurdfdom.cmake

koonpeng:
> I'm afraid that the current implementation is not valid for Windows/MSVC builds.
> 
> Is there any particular reason not to use the urdfdom CMake modules instead of using pkg-config? I think that we could support Windows and other system well by using CMake if possible.

I missed that urdfdom supports cmake. But sdformat prefers the pkgconfig and fallbacks to cmake, should we change it to prefer cmake first? One downside with the cmake config is that it does not support version checking.

j-rivero:
> I missed that urdfdom supports cmake. But sdformat prefers the pkgconfig and fallbacks to cmake, should we change it to prefer cmake first? One downside with the cmake config is that it does not support version checking.

It is my impression that we should prioritize CMake over pkg-config as much as we can. The problem with no version checking with urdfdom CMake is unfortunate and seems a big thing to me. In this case we can document that we try pkg-config first for that reason.

j-rivero:
> > 
> 
> It is my impression that we should prioritize CMake over pkg-config as much as we can. The problem with no version checking with urdfdom CMake is unfortunate and seems a big thing to me. In this case we can document that we try pkg-config first for that reason.

I've implemented the suggestion on d46f7cc4194a99cdceb24783b050edb310750968 branch `jrivero/add-find-urdf` . Let me know if that works for you Koon.

koonpeng:
> I've implemented the suggestion on d46f7cc branch jrivero/add-find-urdf . Let me know if that works for you Koon.

Thanks, I just merged that commit in.

ahcorde:
@osrf-jenkins retest this please

j-rivero (approved):

osrf-triage:

https://community.gazebosim.org/t/new-ignition-releases-2022-01-10/1228/1

PR #196, opened 2021-12-16
Add HIDE_SYMBOLS_BY_DEFAULT option

### LINKED ISSUE #166: Library symbols should be hidden by default
## Environment
* OS Version: Ubuntu 18.04
* Source or binary build? source, ign-cmake2 (17b7058ff5c9de2f6f396d1a000dd0366481d621)

## Description
* Expected behavior: Symbols are hidden by default as stated in https://github.com/ignitionrobotics/ign-cmake/blob/17b7058ff5c9de2f6f396d1a000dd0366481d621/cmake/Export.hh.in#L45-L52
* Actual behavior: Symbols without the `*_HIDDEN` macro are visible

## Steps to reproduce
Here's an MWE: https://github.com/azeey/sandbox/tree/master/ign-cmake-visibility

After building, use `nm -C lib/libtset_ign_cmake_visibility.so | grep test_` to see which symbols are hidden:

```
 nm -C lib/libtest_ign_cmake_visibility.so | grep test_
00000000000005c0 T test_hidden_func(int, int)
00000000000005e0 T test_visible_func(int, int)
00000000000005d0 t test_explicitly_hidden_func(int, int)
```

`test_hidden_func` should be hidden by default, but as the `T` (global symbol) in the `nm` output shows, it is visibile.

scpeters:
# 🎉 New feature

Closes #166

## Summary

Add cmake option to `IgnSetCompilerFlags` that will hide symbols by default if they are not explicitly specified as visible. This option is disabled by default to match the current behavior of ign-cmake.

## Test it

The option is used in https://github.com/ignitionrobotics/sdformat/pull/780/commits/3a37dc8ad62dcefa3f36a696d97e140d6eaedabe from https://github.com/ignitionrobotics/sdformat/pull/780. To confirm that it works, remove some of the `target_sources` cmake calls added in https://github.com/ignitionrobotics/sdformat/pull/780/commits/3a37dc8ad62dcefa3f36a696d97e140d6eaedabe and observe the linking errors.

## Checklist
- [ ] Added tests

### DIFF
--- a/cmake/IgnSetCompilerFlags.cmake
+++ b/cmake/IgnSetCompilerFlags.cmake
@@ -115,11 +115,18 @@ endmacro()
 # Set up compilation flags for GCC or Clang
 macro(ign_setup_gcc_or_clang)
 
+  option(USE_DEFAULT_VISIBILITY_HIDDEN "Hide symbols by default if they are not explicitly specified as visible" FALSE)
+  if(USE_DEFAULT_VISIBILITY_HIDDEN)
+    set(VISIBILITY_FLAG "-fvisibility=hidden")
+  else()
+    set(VISIBILITY_FLAG "-fvisibility")
+  endif()
+
   ign_filter_valid_compiler_options(
     CUSTOM_ALL_FLAGS
         -Wall -Wextra -Wno-long-long -Wno-unused-value -Wfloat-equal
         -Wshadow -Winit-self -Wswitch-default -Wmissing-include-dirs -pedantic
-        -fvisibility)
+        ${VISIBILITY_FLAG})
 
   # -ggdb3: Produce comprehensive debug information that can be utilized by gdb
   set(CUSTOM_DEBUG_FLAGS "-ggdb3")
--- a/cmake/IgnSetCompilerFlags.cmake
+++ b/cmake/IgnSetCompilerFlags.cmake
@@ -115,18 +115,20 @@ endmacro()
 # Set up compilation flags for GCC or Clang
 macro(ign_setup_gcc_or_clang)
 
-  option(USE_DEFAULT_VISIBILITY_HIDDEN "Hide symbols by default if they are not explicitly specified as visible" FALSE)
-  if(USE_DEFAULT_VISIBILITY_HIDDEN)
-    set(VISIBILITY_FLAG "-fvisibility=hidden")
+  if(ign_configure_build_HIDE_SYMBOLS_BY_DEFAULT)
+    set(CMAKE_C_VISIBILITY_PRESET "hidden")
+    set(CMAKE_CXX_VISIBILITY_PRESET "hidden")
   else()
-    set(VISIBILITY_FLAG "-fvisibility")
+    set(CMAKE_C_VISIBILITY_PRESET "default")
+    set(CMAKE_CXX_VISIBILITY_PRESET "default")
   endif()
+
 
   ign_filter_valid_compiler_options(
     CUSTOM_ALL_FLAGS
         -Wall -Wextra -Wno-long-long -Wno-unused-value -Wfloat-equal
         -Wshadow -Winit-self -Wswitch-default -Wmissing-include-dirs -pedantic
-        ${VISIBILITY_FLAG})
+        )
 
   # -ggdb3: Produce comprehensive debug information that can be utilized by gdb
   set(CUSTOM_DEBUG_FLAGS "-ggdb3")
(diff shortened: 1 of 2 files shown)


cmake/IgnSetCompilerFlags.cmake
```diff
@@ -115,11 +115,18 @@ endmacro()
 # Set up compilation flags for GCC or Clang
 macro(ign_setup_gcc_or_clang)
 
+  option(USE_DEFAULT_VISIBILITY_HIDDEN "Hide symbols by default if they are not explicitly specified as visible" FALSE)
```
azeey, line 118: I'm getting this error:
```

CMake Warning (dev) at /home/addisu/ws/dome/install/share/cmake/ignition-cmake2/cmake2/IgnSetCompilerFlags.cmake:118 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  For compatibility with older versions of CMake, option is clearing the
  normal variable 'USE_DEFAULT_VISIBILITY_HIDDEN'.
  /home/addisu/ws/dome/install/share/cmake/ignition-cmake2/cmake2/IgnSetCompilerFlags.cmake:59 (ign_setup_gcc_or_clang)
  /home/addisu/ws/dome/install/share/cmake/ignition-cmake2/cmake2/IgnConfigureBuild.cmake:86 (ign_set_compiler_flags)
  CMakeLists.txt:92 (ign_configure_build)
This warning is for project developers.  Use -Wno-dev to suppress it.

```

scpeters: what version of cmake are you using? and what OS?

azeey: Ah, I should have mentioned that. I'm on Ubuntu 20.04, cmake 3.16.3

cmake/IgnSetCompilerFlags.cmake
```diff
@@ -115,11 +115,18 @@ endmacro()
...
+  else()
+    set(VISIBILITY_FLAG "-fvisibility")
+  endif()
+
   ign_filter_valid_compiler_options(
     CUSTOM_ALL_FLAGS
         -Wall -Wextra -Wno-long-long -Wno-unused-value -Wfloat-equal
         -Wshadow -Winit-self -Wswitch-default -Wmissing-include-dirs -pedantic
-        -fvisibility)
+        ${VISIBILITY_FLAG})
```
koonpeng, line 129: cmake has a visibility property https://cmake.org/cmake/help/v3.16/prop_tgt/LANG_VISIBILITY_PRESET.html, is there a way to make use of that?

cmake/IgnSetCompilerFlags.cmake
koonpeng, line 118: I think we should make this on by default in the next major version. I think many of other ignition packages also assume it's hidden by default, since they are using `_VISIBLE` macros instead of `_HIDDEN` macros.

cmake/IgnSetCompilerFlags.cmake
scpeters: good call!

should we also set [VISIBILITY_INLINES_HIDDEN](https://cmake.org/cmake/help/v3.16/prop_tgt/VISIBILITY_INLINES_HIDDEN.html#prop_tgt:VISIBILITY_INLINES_HIDDEN)?

scpeters: we are now setting `CMAKE_C_VISIBILITY_PRESET` and `CMAKE_CXX_VISIBILITY_PRESET` as of https://github.com/ignitionrobotics/ign-cmake/pull/196/commits/fd38458199dd12efb2239c7853159a0096f6e55b

cmake/IgnSetCompilerFlags.cmake
scpeters: I changed it to an argument to `ign_configure_build` instead of an `option` in https://github.com/ignitionrobotics/ign-cmake/pull/196/commits/fd38458199dd12efb2239c7853159a0096f6e55b. I think that resolves the warning

azeey: Yup, that resolved it for me. Thanks!

azeey: Agreed. I don't think we should close #166 until it becomes the default.

azeey (approved):
Works great! I've used `abidiff` to compared the shared libraries of ign-math and sdformat created before and after this PR and they were no differences.

osrf-triage:

https://community.gazebosim.org/t/new-ignition-releases-2022-01-10/1228/1

PR #203, opened 2021-12-29
Improve static library support

scpeters:
# 🦟 Bug fix

Port of ignitionrobotics/sdformat#394 from @joxoby to ign-cmake2. This publicly defines the `*_STATIC_DEFINE` symbol for a library target if `BUILD_SHARED_LIBS` is Off.

## Summary

I believe a special symbol needs to be defined when building with static libraries on Windows. Logic for defining this symbol was added to libsdformat11 in https://github.com/ignitionrobotics/sdformat/pull/394, and I noticed it while preparing to forward-port https://github.com/ignitionrobotics/sdformat/pull/780 to libsdformat11, since I noticed that ign-cmake2 doesn't currently define this symbol. I've added some examples using static libraries, but I'm not sure they're rich enough to prove this fixes any problems on Windows. Perhaps @traversaro @joxoby @madebr have some suggestions?

## Checklist
- [X] Added tests
- [X] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [X] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))

### DIFF
--- a/cmake/IgnUtils.cmake
+++ b/cmake/IgnUtils.cmake
@@ -1416,6 +1416,10 @@ macro(_ign_add_library_or_component)
     set(install_include_dir
       "${IGN_INCLUDE_INSTALL_DIR_FULL}/${include_dir}")
 
+    if (NOT BUILD_SHARED_LIBS)
+      target_compile_definitions(${lib_name} PUBLIC ${export_base}_STATIC_DEFINE)
+    endif()
+
     # Configure the installation of the automatically generated file.
     install(
       FILES "${implementation_file_name}"
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -13,8 +13,10 @@ set(example_directories
   no_ignition_prefix
   prerelease
   core_nodep
+  core_nodep_static
   core_child
   core_child_private
+  core_static_child
   comp_deps
   )
 if (NOT CMAKE_GENERATOR MATCHES "Visual Studio")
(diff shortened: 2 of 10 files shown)


traversaro:
I am currently away from laptop keyboard, but if I recall correctly this should not be necessary as the static case is already handled by the generate_export_header macro. Manually defining the static macro is only necessary when using the same header for both shared and static variants of a library (see https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html) or when using an hardcoded export header file (as we were doing in sdformat if I recall correctly).

traversaro:
I guess also @Ace314159 could be interested.

traversaro:
> I am currently away from laptop keyboard, but if I recall correctly this should not be necessary as the static case is already handled by the generate_export_header macro.

Related code: https://github.com/Kitware/CMake/blob/master/Modules/GenerateExportHeader.cmake#L302 .

scpeters:
> I am currently away from laptop keyboard, but if I recall correctly this should not be necessary as the static case is already handled by the generate_export_header macro. Manually defining the static macro is only necessary when using the same header for both shared and static variants of a library (see https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html) or when using an hardcoded export header file (as we were doing in sdformat if I recall correctly).

ah, I believe you are right. I had noticed the `*_STATIC_DEFINE` macros, but didn't realize that it would generate the macros differently for static libraries. I will revert https://github.com/ignitionrobotics/ign-cmake/pull/203/commits/9666346ffd8cea16ef6c1a40a5cb5c48b90a3450 and just keep the new examples. I'll also revert https://github.com/ignitionrobotics/sdformat/pull/808/commits/72c4853f1dc5a1240b04b85bbfb0304b1450901b from https://github.com/ignitionrobotics/sdformat/pull/808, since I think it won't be needed either

scpeters:
examples only in https://github.com/ignitionrobotics/ign-cmake/pull/202

PR #213, opened 2022-03-19
Replace deprecated PythonInterp with Python3

### LINKED ISSUE #212: Update CMake Python code to only find Python3
## Description
The [current approach](https://github.com/ignitionrobotics/ign-cmake/blob/ign-cmake2/cmake/IgnPython.cmake) to find Python is [deprecated since CMake 3.12](https://cmake.org/cmake/help/latest/module/FindPythonInterp.html). I also think that all Python scripts need Python3 anyway (Python2 is EOL) but [Ubuntu Bionic ships with CMake 3.10](https://packages.ubuntu.com/bionic/cmake) so a fallback is needed until it is also EOL.
My suggestion is to use the following
```cmake
if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12")
    find_package(Python3)
else()
    set(PYTHON_VERSION "" CACHE STRING
    "Specify specific Python version to use ('major.minor' or 'major')")

    # if not specified otherwise use Python 3
    if(NOT PYTHON_VERSION)
    set(PYTHON_VERSION "3")
    endif()

    find_package(PythonInterp ${PYTHON_VERSION})
endif()
```
and as soon as Ubuntu Bionic reaches the EOL in April 2023 it can be removed and the [minimum CMake version be raised to 3.16](https://packages.ubuntu.com/focal/cmake).
Doing the change might need some adjustments between the different projects as e.g. `${PYTHON_VERSION}` will be `${Python3_VERSION}` in case the variables are used somewhere. 

However, there are also certain cases (e.g. [ign-math gtest](https://github.com/ignitionrobotics/ign-math/blob/ign-math6/test/gtest/cmake/internal_utils.cmake#L225), [ign-rendering gtest](https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering6/test/gtest/cmake/internal_utils.cmake#L225) and probably in every other gtest) where the code is "duplicated" instead of using the ign-cmake include.

Thus I created this issue to discuss the change before looking into each detail and making the changes.

Bi0T1N:
# 🦟 Bug fix

Fixes #212 

## Summary
`PythonInterp` is deprecated for CMake versions > 3.11 and all scripts use Python3 features anyway, so replace it with the recommended `find_package(Python3)` but also provide fallback solutions until the required CMake version can be bumped.
**Note**: Some small adjustments may be required in other packages to use the new variables. If the changes here are fine I will create the other PR's.

## Checklist
- [ ] Added tests

### DIFF
--- a/cmake/IgnPython.cmake
+++ b/cmake/IgnPython.cmake
@@ -12,14 +12,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Copied from ament/ament_cmake: ament_cmake/ament_cmake_core/cmake/core/python.cmake
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.19")
+  set(PYTHON_VERSION "" CACHE STRING
+    "Specify specific Python3 version to use ('major.minor' or 'versionMin...[<]versionMax')")
 
-set(PYTHON_VERSION "" CACHE STRING
-  "Specify specific Python version to use ('major.minor' or 'major')")
+  find_package(Python3 ${PYTHON_VERSION} QUIET)
+elseif(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12")
+  # no support for finding specific versions
+  find_package(Python3 QUIET)
+else()
+  # TODO: remove this block as soon as the CMake version can safely be bumped to => 3.12
+  set(PYTHON_VERSION "" CACHE STRING
+    "Specify specific Python version to use ('major.minor' or 'major')")
 
-# if not specified otherwise use Python 3
-if(NOT PYTHON_VERSION)
-  set(PYTHON_VERSION "3")
+  # if not specified otherwise use Python 3
+  if(NOT PYTHON_VERSION)
+    set(PYTHON_VERSION "3")
+  endif()
+
+  find_package(PythonInterp ${PYTHON_VERSION} QUIET)
+
+  if(PYTHONINTERP_FOUND)
+    set(Python3_Interpreter_FOUND ${PYTHONINTERP_FOUND})
+    set(Python3_EXECUTABLE ${PYTHON_EXECUTABLE})
+  endif()
 endif()
-
-find_package(PythonInterp ${PYTHON_VERSION} QUIET)
--- a/cmake/IgnPython.cmake
+++ b/cmake/IgnPython.cmake
@@ -13,24 +13,24 @@
 # limitations under the License.
 
 if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.19")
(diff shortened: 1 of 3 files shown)


j-rivero:
Thanks for the PR @Bi0T1N! Since the changes are carefully maintaining current behavior I'm going to target the PR against the ign-cmake2 branch so we get the changes ready in a released branch, ready to be shipped in the next release.

j-rivero:
> Thanks for the PR @Bi0T1N! Since the changes are carefully maintaining current behavior I'm going to target the PR against the ign-cmake2 branch so we get the changes ready in a released branch, ready to be shipped in the next release.

Change base in the PR did not work well, keeping it against main. We'll do the other way around, We can backport it to `ign-cmake2`.

cmake/IgnPython.cmake
```diff
@@ -12,14 +12,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Copied from ament/ament_cmake: ament_cmake/ament_cmake_core/cmake/core/python.cmake
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.19")
+  set(PYTHON_VERSION "" CACHE STRING
```
j-rivero, line 16: Could we please rename `PYTHON_VERSION` to avoid possible conflicts using that generic name? Maybe using `IGN_PYTHON_VERSION`?

cmake/IgnPython.cmake
```diff
@@ -12,14 +12,28 @@
...
+    "Specify specific Python3 version to use ('major.minor' or 'versionMin...[<]versionMax')")
 
-set(PYTHON_VERSION "" CACHE STRING
-  "Specify specific Python version to use ('major.minor' or 'major')")
+  find_package(Python3 ${PYTHON_VERSION} QUIET)
+elseif(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12")
+  # no support for finding specific versions
+  find_package(Python3 QUIET)
```
j-rivero, line 22: I was thinking on warning the user if  `PYTHON_VERSION` is being set but CMake is already doing it for using of CMake lower than 3.19 and greater than 3.12 (like Focal).

j-rivero (commented):
I've been testing Ignition Fortress and Ignition Garden packages with this and seems to work nicely in both Ubuntu Focal (version 3.16) and Debian Sid (version 3.22). 

I left one suggestion to change the input name of the variable, otherwise we are good to go.

cmake/IgnPython.cmake
Bi0T1N: Done :)

j-rivero (approved):
Great work @Bi0T1N . Good to go.

PR #217, opened 2022-03-29
Add "-Wconversion" to more closely align to MSVC

mjcarroll:
This will make GCC builds warn on the same things that Windows does.  Unfortunately, it will also warn on more things, so I'm going to open this as a draft until the whole stack builds cleanly.

-  ign-utils: https://github.com/ignitionrobotics/ign-utils/pull/46

### DIFF
--- a/cmake/IgnSetCompilerFlags.cmake
+++ b/cmake/IgnSetCompilerFlags.cmake
@@ -126,7 +126,7 @@ macro(ign_setup_gcc_or_clang)
 
   ign_filter_valid_compiler_options(
     CUSTOM_ALL_FLAGS
-        -Wall -Wextra -Wno-long-long -Wno-unused-value -Wfloat-equal
+        -Wall -Wextra -Wno-long-long -Wno-unused-value -Wfloat-equal -Wconversion
         -Wshadow -Winit-self -Wswitch-default -Wmissing-include-dirs -pedantic
         )
 


j-rivero:
The level of strictness and the level of warnings that we want to enable by default that we want to have with ign-cmake is open to debate however I have some thoughts about this PR:
  - We need to define clearly what are we trying to get with the change: try to unify compiler warnings behaviour? improve the QA of the code using ign-cmake by increasing the level of strictness?
 - Enabling the warning flag by default on ign-cmake2 branch is probably a no-go since it is going to alter many builds out there by introducing new warnings.

Depending on the first point, we can try to generalize the measure into new features. Seems to me like we are going to need to control new behavior using a CMake parameter (or similar) no matter what options we choose.

mjcarroll:
> We need to define clearly what are we trying to get with the change: try to unify compiler warnings behaviour?

I was mainly trying to bring our linux gcc warnings in line with what MSVC produces in terms of numeric conversions.  Unfortunately, after working up the stack a little bit, GCC is _far_ more strict (to the tune of dozens of warnings on ign-math itself).  I think this is probably a non-starter as default flag for the reason you said.

I'm going to leave this as a draft just so that I can track progress towards being able to enable it, but I agree it will need to be changed to an opt-in mechanism rather than opt-out.

PR #225, opened 2022-04-05
Macros for generating cmd*.rb , conf/*.yaml files

harshmahesheka:
# 🎉 New feature

Closes https://github.com/ignitionrobotics/ign-tools/issues/72

## Summary
Added Macros for generating  cmd*.rb and conf/*.yaml files which decreases boilerplate code from repositories 

## Checklist
- [ ] Added tests

### DIFF
--- a/cmake/IgnUtils.cmake
+++ b/cmake/IgnUtils.cmake
@@ -1791,3 +1791,75 @@ macro(_ign_cmake_parse_arguments prefix options oneValueArgs multiValueArgs)
   endif()
 
 endmacro()
+
+#################################################
+# ign_generate_conf()
+# Generate conf/*.yaml files for projects
+macro(ign_generate_conf)
+    # Used only for internal testing.
+    set(ign_library_path "${CMAKE_BINARY_DIR}/test/lib/$<CONFIG>/ruby/ignition/cmd${IGN_DESIGNATION}${PROJECT_VERSION_MAJOR}")
+
+    # Generate a configuration file for internal testing.
+    # Note that the major version of the library is included in the name.
+    configure_file(
+      "${IGN_DESIGNATION}.yaml.in"
+      "${CMAKE_CURRENT_BINARY_DIR}/${IGN_DESIGNATION}${PROJECT_VERSION_MAJOR}.yaml.configured" @ONLY)
+
+    file(GENERATE
+      OUTPUT "${CMAKE_BINARY_DIR}/test/conf/$<CONFIG>/${IGN_DESIGNATION}${PROJECT_VERSION_MAJOR}.yaml"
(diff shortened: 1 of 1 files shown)


j-rivero:
Thanks Harsh for PR! While ign-cmake can sound like the best place to put our CMake implementations related to Ignition Gazebo, in this case I see we are adding specific functions that will only work with the software `ign-tools`. Sounds to me like a more natural approach to implement them in that package (ign-tools CMake module) instead of here. @scpeters @adlarkin, what do you think?

scpeters:
> Thanks Harsh for PR! While ign-cmake can sound like the best place to put our CMake implementations related to Ignition Gazebo, in this case I see we are adding specific functions that will only work with the software `ign-tools`. Sounds to me like a more natural approach to implement them in that package (ign-tools CMake module) instead of here. @scpeters @adlarkin, what do you think?

I agree with installing it with `ign-tools`, as I suggested in the original issue https://github.com/ignitionrobotics/ign-tools/issues/72:

> We could install that with ign-tools since it is related to this package.

harshmahesheka:
> > Thanks Harsh for PR! While ign-cmake can sound like the best place to put our CMake implementations related to Ignition Gazebo, in this case I see we are adding specific functions that will only work with the software `ign-tools`. Sounds to me like a more natural approach to implement them in that package (ign-tools CMake module) instead of here. @scpeters @adlarkin, what do you think?
> 
> I agree with installing it with `ign-tools`, as I suggested in the original issue [ignitionrobotics/ign-tools#72](https://github.com/ignitionrobotics/ign-tools/issues/72):
> 
> > We could install that with ign-tools since it is related to this package.

I thought it is a cmake macro so, ign-cmake would be a better place for it but yes you are correct it is a specific function and ign-transport would be a more suitable place for it. So, should I close this pull request and create a new one in ign-transport?

scpeters:
> > > Thanks Harsh for PR! While ign-cmake can sound like the best place to put our CMake implementations related to Ignition Gazebo, in this case I see we are adding specific functions that will only work with the software `ign-tools`. Sounds to me like a more natural approach to implement them in that package (ign-tools CMake module) instead of here. @scpeters @adlarkin, what do you think?
> > 
> > 
> > I agree with installing it with `ign-tools`, as I suggested in the original issue [ignitionrobotics/ign-tools#72](https://github.com/ignitionrobotics/ign-tools/issues/72):
> > > We could install that with ign-tools since it is related to this package.
> 
> I thought it is a cmake macro so, ign-cmake would be a better place for it but yes you are correct it is a specific function and ign-transport would be a more suitable place for it. So, should I close this pull request and create a new one in ign-transport?

I think `ign-tools` would be a better place than `ign-transport`

j-rivero:
> Sorry, I mistyped. So,should I add this to ign-tools?

That would be great, thanks! Feel free to add me to reviewers there or just mention my user so I review the code there.

PR #230, opened 2022-04-09
Adding tests for CONFIG argument

harshmahesheka:
# 🎉 New feature

Requires  https://github.com/ignitionrobotics/ign-cmake/pull/211
## Summary
Adds tests for CONFIG argument in ign_find_packages().The idea being `ign_find_package(ignition-find_config) `will search for `Findignition-find_config.cmake` before `ignition-find_configConfig.cmake ` in MODULE mode whereas when we pass CONFIG option it will skip to CONFIG mode and directly search for` ignition-find_configConfig.cmake`.
## Checklist
- [ ] Added tests


examples/use_config_ifp/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+find_package(ignition-cmake2 REQUIRED)
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/")
+set(ignition-find_config_DIR "${CMAKE_SOURCE_DIR}/")
+#ign_find_package(ignition-find_config) will search for Findignition-find_config.cmake before ignition-find_configConfig.cmake 
+ign_find_package(ignition-find_config)
+#After CONFIG arguemnt ign_find_package will skip MODULE mode and search for ignition-find_configConfig.cmake directly
+ign_find_package(ignition-find_config CONFIG)
+ign_configure_project()
+ign_configure_build(QUIT_IF_BUILD_ERRORS)
+ign_create_packages()
```
adlarkin, line 12: It looks like these lines are making CI angry, I wonder if we should remove them. I see the following error output when I try to build this example locally:
```
CMake Error at /usr/local/share/cmake/ignition-cmake2/cmake2/IgnConfigureBuild.cmake:142 (add_subdirectory):
  add_subdirectory given source "src" which is not an existing directory.
  CMakeLists.txt:11 (ign_configure_build)

...

-- Configuring incomplete, errors occurred!
```

I did notice that earlier in the output, `CONFIG` is working as expected, because I see the following messages:
```
-- The package was found using MODULE mode

...

-- The package was found using CONFIG mode
```

So, since this test is only meant to test `CONFIG` and not configuring the project/build, I'm leaning towards removing these lines since the `CONFIG` functionality that's being tested is working. What do you think, @j-rivero?

examples/use_config_ifp/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+
```
adlarkin, line 13: nit: this empty line can be deleted, we typically don't leave whitespace at the end of files

examples/use_config_ifp/CMakeLists.txt
harshmahesheka: Hey @adlarkin, adding an src folder with an empty CMakeLists.txt and an empty README.md seems to have solved the problem. Also, I have deleted all the unnecessary whitespaces.

adlarkin: Great, thanks! Everything looks good to me :slightly_smiling_face:

adlarkin (approved):

scpeters (commented):
it looks like the new tests print console messages so you can see interactively if it is working, but not otherwise. It would be nice if the test would fail if the wrong cmake file was found, perhaps by setting variables and then expecting them to have a certain value and reporting a cmake error if the expected value was not found, but that can be future work.

Thanks for your contribution!

harshmahesheka:
> expecting them to have a certain value and reporting a cmake error if the expected value was not found, but that can be future work.

Yeah, we can do it this way also. Should I make the changes and open a pr?

osrf-triage:

https://community.gazebosim.org/t/new-releases-2022-04-27-fortress-citadel/1389/1

Issue #171, opened 2021-05-14
Deprecate utilities in favor of ign-utils


chapulina:
The [ign-utils](https://github.com/ignitionrobotics/ign-utils/) package has been created to hold general purpose utilities that have minimal dependencies and are useful for multiple libraries. We've already moved the [SuppressWarning](https://github.com/ignitionrobotics/ign-utils/blob/ign-utils1/include/ignition/utils/SuppressWarning.hh) functionality there.

I think we should also move the `ExtraTestMacros` and to `ign-utils` and deprecate the `utilities` component of this package. I can see an argument for keeping that here close to other test configurations though. My main concern is that this repository is not well suited to hold C++ code. We have created a new namespace, `utilities`, which doesn't match the rest of this library. We aren't generating test coverage for it, etc. I think it's a cleaner separation of concerns if we leave C++ code out of `ign-cmake`.

## Desired behavior

On `ign-cmake3`, using `ignition::utilities` generates deprecation warnings. On `ign-cmake4`, the component is removed.

I suggest this with the full understanding that `ign-cmake4` may only be released in 2078 :upside_down_face: 

My short-term suggestion is to migrate these utilities to `ign-utils` and stop using the ones from `ign-cmake`, even though they're not officially deprecated on `ign-cmake2`.

## Alternatives considered

We could just keep `ExtraTestMacros` here... That would certainly involve less effort. My main concern is that we forget the separation of concerns and start adding more header-only functionality here, when it should go to `ign-utils`.

## Implementation suggestion

1. Add `ExtraTestMacros` to `ign-utils`
2. Migrate all uses of `utilities` to use `ign-utils`
3. Remove all dependencies on the `utilities` component from Fortress
4. Initiate the tick-tock of utilities suggested above.

## Additional context

See #18 for why the utilities were added in the first place.

mjcarroll:
This copies `ExtraTestMacros` into ign-utils: https://github.com/ignitionrobotics/ign-utils/pull/37

chapulina:
Fully deprecated as of #233

PR #243, opened 2022-04-27
ign -> gz Header Migration : ign-cmake

methylDragon:
See https://github.com/ignition-tooling/release-tools/issues/711


methylDragon:
# Note
This PR will require a lot of manual patching for `REPLACE_IGNITION_INCLUDE_PATH` for the examples because the examples have a bunch of subdirectories/subprojects.

scpeters:
> # Note
> This PR will require a lot of manual patching for `REPLACE_IGNITION_INCLUDE_PATH` for the examples because the examples have a bunch of subdirectories/subprojects.
> 
> I won't script that because there's a lot of variance between how the examples are done for the different packages.

I think the examples in this repository can wait if they take extra work. They can switch when we update the default include path in the ign-cmake macros themselves

methylDragon:
@osrf-jenkins run tests please!

include/gz/utilities/ExtraTestMacros.hh
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ *
+ */
+
+#ifndef GZ_UTILITIES_EXTRATESTMACROS_HH
+#define GZ_UTILITIES_EXTRATESTMACROS_HH
+
+#include <gz/utilities/detail/ExtraTestMacros.hh>
+
+#pragma message("ign-cmake (utilities) ExtraTestMacros is deprecated, use ign-utils")
```
chapulina, line 23: Since this is deprecated anyway, how about we don't add `gz` aliases for them? That is, users of `ignition/utilities` shouldn't migrate to `gz/utilities`, but to `gz/utils`, which is in another library.

chapulina (changes requested):
Looking at this PR now, I think `gz-cmake` is a special case where we don't need to migrate any includes to `gz` because they're all deprecated (#171). So I think we can close this PR and call the `gz-cmake` header migration done.

methylDragon:
> Looking at this PR now, I think `gz-cmake` is a special case where we don't need to migrate any includes to `gz` because they're all deprecated (#171). So I think we can close this PR and call the `gz-cmake` header migration done.

:thinking: We might want to consider migrating the examples when we're ready then, but I'll close this for now.

scpeters:
> > Looking at this PR now, I think `gz-cmake` is a special case where we don't need to migrate any includes to `gz` because they're all deprecated (#171). So I think we can close this PR and call the `gz-cmake` header migration done.
> 
> 🤔 We might want to consider migrating the examples when we're ready then, but I'll close this for now.

we can migrate the `#include <ignition/*` statements in the examples when we change the default `PROJECT_INCLUDE_PATH` path to `gz/`

* https://github.com/ignitionrobotics/ign-cmake/blob/fd2e2610d9138dad8cc3214d4f00ddf98ab3b599/cmake/IgnConfigureProject.cmake#L91

PR #245, opened 2022-05-03
[ign ➡️  gz] Internal parsed variables

chapulina:
* Part of https://github.com/ignition-tooling/release-tools/issues/719

## Summary

This PR updates just the variables parsed by `cmake_parse_arguments` to use `gz` instead of `ign`. Those variables are internal to each function and shouldn't affect the public API.

⬅️ This PR can be backported to `ign-cmake2` later to reduce the diff between the branches.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [x] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸

PR #247, opened 2022-05-03
[ign ➡️  gz] Internal functions

chapulina:
* Builds on top of #245 
* Part of https://github.com/ignition-tooling/release-tools/issues/719

## Summary

Migrate functions that are internal to `gz-cmake` to use `gz` instead of `ign`. I considered private functions:

* Functions that have documentation saying that they're private
* Functions that are prefixed by `_`, even if their docs don't say they're private - I'm relying on convention here.
* A few other functions that I didn't see being used by other libraries. I updated the docs to mention they're private.

It is possible that downstream projects are relying on these functions and will be broken, but that's unlikely. I'd like to reduce our API surface by making as many functions private as possible. This way we can focus our tick-tock efforts into functions that we know are used by downstream projects.

## Checklist
- [ ] Added tests


methylDragon:
I'm wondering if it'll make sense to also make function/macro-local variables explicitly private, both:
- because in the case of macros, they'll bubble up to the parent scope
- clarity when editing the function/macro body

:O

chapulina:
> make function/macro-local variables explicitly private

I tried to do that using the `_` prefix, but that's just convention. Is there a way to enforce a private scope at the CMake level?

methylDragon:
> > make function/macro-local variables explicitly private
> 
> I tried to do that using the `_` prefix, but that's just convention. Is there a way to enforce a private scope at the CMake level?

Because CMake macros do text substitution, not really.. unless you define a function inside the macro (with the macro arg as part of the function name) and use that function. But at that point you might as well use a function :grimacing:

Converting macros to functions naively can have follow on consequences though, because macro variables are parent scoped, but function variables are not, which, for this library at least, is probably a lot of work, and not too much of a priority imho. The `_` convention is fine then :thinking:

methylDragon (approved):
> > The _ convention is fine then
> 
> I'll take this to mean that you're not requesting any changes from this PR, @methylDragon . Let me know if you find any issues with it.

Ah yes!

I think we're good to go?

PR #251, opened 2022-05-06
Add LTCG flag on Windows builds

j-rivero:
# 🦟 Bug fix

## Summary

Include LTCG flags as suggested by the own compiler since we are using GL


mjcarroll (approved):

mjcarroll:
Just wanted to add a note before merging:

This flag shouldn't be changing any behavior, as our builds already have this warning in it from enabling `/GL`

```
[33.391s]   Environment.obj : MSIL .netmodule or module compiled with /GL found; restarting link with /LTCG; add /LTCG to the link command line to improve linker performance
```

Additionally, the documentation for `/LTCG` says that it is implied for `/GL`: https://docs.microsoft.com/en-us/cpp/build/reference/ltcg-link-time-code-generation?view=msvc-170

Worst case is that this is a no-op, best case is that it cleans up some warnings.

PR #273, opened 2022-07-09
ign -> gz FindXXX migration : gz-cmake

methylDragon:

# NOTE: THIS MUST BE REBASE MERGED!! Files were moved!

# Description
All `FindXXX.cmake` files have been hard migrated to Gz.

Some downstream packages still use them via the `gz_find_package()` call, so an intercepting redirection of any `Ign` prefixed find calls to `Gz` has been implemented (with warning.)

image

# Hard-tocks
 - `<Package_name>_FIND_VERSION_MINOR` CMake variables are being used (but populated automatically from the `find_package()` call, which might be the Gz or Ign version
  - Hard-tocked these since the file names changed (and the only way they'll be found is if the redirected find_package() call uses the Gz prefixed package name, which will set the Gz prefixed `_FIND_VERSION_MINOR` var)
- Likewise with `XXX_FIND_COMPONENTS`
  - These are internal to the best of my knowledge, so the user shouldn't be using them

# Tick-tocks
- Downstream libs are using XXX_FOUND variables, I will ticktock those
- Explicitly being `set()` with an Ign prefix, just in case they'll be used downstream

# Notes
- `gz_import_target` is being called to generate targets.

The call itself is made of several components:
```
  gz_import_target(IgnOGRE2
    TARGET_NAME IgnOGRE2::IgnOGRE2
    LIB_VAR OGRE2_LIBRARIES
    INCLUDE_VAR OGRE2_INCLUDE_DIRS)
```
That is,
```
  gz_import_target(<package_name>
    TARGET_NAME <TARGET>::<TARGET>
    LIB_VAR OGRE2_LIBRARIES
    INCLUDE_VAR OGRE2_INCLUDE_DIRS)
```

- The package name is hard-tocked, and library aliases are installed for the imported target name.
- In order to add alias targets, however, I needed to set the imported `add_library()` call to `GLOBAL` visibility.
  - https://github.com/gazebosim/gz-cmake/pull/273/commits/c409c94fc5f1bbf64f56b73329104e927fb438bd#diff-42bab6079b4ba6149374a4f37a702aabec4cd82b534287747b8014a4a1a39eb8R110
  - I don't know what implications this will have on performance, but it works. To be safe we should remove the GLOBAL keyword once we don't need the tick-tock.


scpeters:
testing sdformat against this branch in macOS and windows CI using the `ci_matching_branch/` trick in https://github.com/gazebosim/sdformat/pull/1072

cmake/IgnUtils.cmake
```diff
@@ -209,7 +210,37 @@ macro(gz_find_package PACKAGE_NAME)
...
+
+      set(${PACKAGE_NAME_GZ}_find_package_args ${${PACKAGE_NAME}_find_package_args})
+      list(POP_FRONT ${PACKAGE_NAME_GZ}_find_package_args)
+      list(PREPEND ${PACKAGE_NAME_GZ}_find_package_args ${PACKAGE_NAME_GZ})
+
+      find_package(${${PACKAGE_NAME_GZ}_find_package_args})
+      if(${PACKAGE_NAME_GZ}_FOUND)
+
+        message(DEPRECATION "Ign prefixed package name [${PACKAGE_NAME}] is deprecated! Automatically using the Gz prefix instead: [${PACKAGE_NAME_GZ}]")
+        set(PACKAGE_NAME ${PACKAGE_NAME_GZ})
```
scpeters, line 232: this change in `PACKAGE_NAME` means there is no longer a matching `${PACKAGE_NAME}_pretty` variable, since that is set just before this code block. This could be fixed by moving that code block to after this `if` logic

methylDragon: Ah, nice catch! It seems despite my best efforts I got bamboozled by macro substitution :sweat: 

https://github.com/gazebosim/gz-cmake/pull/273/commits/98aadc506549c33e228051dee7cfeebc35cf3bcf

Upon approval, I'll squash all the non file-migration commits into one so when we rebase merge we'll just have two commits!

scpeters (approved):

methylDragon:
Gonna squash all the non moving commits :eyes:

PR #274, opened 2022-07-09
ign -> gz CMake, Config, and Example Files migration : gz-cmake

methylDragon:
See: gazebo-tooling/release-tools#698

- Needs merge of https://github.com/gazebosim/gz-cmake/pull/273

# NOTE: THIS MUST BE REBASE MERGED!! Files were moved!

# Description
Migrates IgnXXX.cmake files, examples, and configured files.

# Hard-tocks
- ignition.*.in and ign.*.in files have been hard migrated since they're used internally. References to them in `gz-cmake` have also been migrated.

# Tick-tocks
- All `IgnXXX.cmake` files have been migrated to Gz.

Some downstream packages use `IgnPython.cmake` and `IgnBenchmark.cmake`, symlinks have been installed for tick-tocking (though warnings cannot be emitted.)

PR #275, opened 2022-07-09
ign -> gz Migrate Internal CMake Vars : gz-cmake

methylDragon:

Built on top of: https://github.com/gazebosim/gz-cmake/pull/274

In an effort to keep track of what has already been tick-tocked, I placed the `Deprecated.` comment (in most cases) on the same line so it appears on grep.

Try it: `grep -ri set\(ign`

# BREAKING NOTES
- `IGNITION_DOXYGEN_TAGFILES` is set by DOWNSTREAM libraries and used in this UPSTREAM library... I am just going to hard-tock it and remember to update the downstream usage since it's just for docs...
  - This won't lead to a compile error, but will lead to malformed docs until the downstream libraries are tocked.

The next few ones look like they're breaking (because they change the install path to `gz`), **but local builds work**.
Release repos might need to be updated to have the topline `ignition` -> `gz`.

- Change default include directory from `include/ignition` to `include/gz` (I think in `GzConfigureBuild.cmake` and `GzPackaging.cmake`?)
- `PROJECT_INCLUDE_DIR` stores `ignition/${IGN_DESIGNATION}` -> `gz/...`
  - This would be breaking, but all downstream uses actually bypass this block, so I think it's safe to adjust, I think
  - This crucially changes the `install` directory to `gz` instead of `ignition`, likely necessitating release repository changes, but no other issues if packages are rebuilt accordingly

See: https://github.com/gazebosim/gz-cmake/pull/275/commits/e1c4a355353937c1c11cb2841356e09f348051b2

# Pending??
- [What should be done about `version_info.json.in`](https://github.com/gazebosim/gz-cmake/blob/main/cmake/version_info.json.in#L40) (I've provisionally migrated `ignition` to `gz` here)

# Hard-Tocks
Usually because they're internal or aren't used anywhere else in the entire Gz stack
- GitHub workflow targets

- Erm.. docs only mentions?
  - `IGN_SANITIZERS`
  - `USE_IGN_RECOMMENDED_FLAGS`
- `GzCodeCoverage.cmake`:
  - `IGN_SETUP_TARGET_FOR_COVERAGE`
- Top level gz-cmake CMakeLists.txt
  - `ign_pkgconfig_xxx`
  - `ign_utilities_xxx`
  - `IGN_PC_CONFIG_RELATIVE_PATH_TO_PREFIX`
- `GzPackaging.cmake`:
  - `IGNITION_CMAKE_DIR`
  - `IGNITION_CMAKE_VERSION_MAJOR`
- `GzCreateDocs.cmake`
  - `IGNITION_DOXYGEN_GENTAGFILE`
  - `IGNITION_DOXYGEN_INPUT`
  - `IGNITION_DOXYGEN_IMAGE_PATH`
  - `IGNITION_DOXYGEN_API_MAINPAGE_MD`
  - `IGNITION_DOXYGEN_GENHTML`
  - `IGNITION_DOXYGEN_AUTOGENERATED_DOC`
  - `IGNITION_DOXYGEN_TUTORIALS_MAINPAGE_MD`
  - `IGNITION_DOXYGEN_TUTORIALS_DIR`
  - `IGNITION_DOXYGEN_ADDITIONAL_INPUT_DIRS`
- `GzRonn2Man.cmake`
  - `ign_add_manpage_target`

- `gz-config.cmake.in`
  - `set_and_check(@PKG_NAME@_INCLUDE_DIRS "@PACKAGE_IGN_INCLUDE_INSTALL_DIR_FULL@")` this is the only occurance of that IGN variable. I don't know why

# Tick-Tocks
- Every variable that is set with an IGNITION/IGN prefix, generally in `.cmake` files which are meant to be included (internal references hard-tocked, but the set call itself is ticktocked.)

- `gz-cmake-config.cmake.in`:
  - `IGN_INCLUDE_INSTALL_DIR_FULL`
  - `IGNITION_CMAKE_VERSION_MAJOR`
  - `IGNITION_CMAKE_DOXYGEN_DIR` (Also in top level CMakeLists.txt... it's confusing, I know)
  - `IGNITION_CMAKE_CODECHECK_DIR`
  - `IGNITION_CMAKE_BENCHMARK_DIR`
  - `IGNITION_CMAKE_TOOLS_DIR`

- All these IGN_DESIGNATION variables...
  - `IGN_DESIGNATION`
  - `IGN_DESIGNATION_LOWER`
  - `IGN_DESIGNATION_UPPER`
  - `IGN_DESIGNATION_FIRST_LETTER`
  - `IGN_DESIGNATION_CAP`

- `GzPackaging.cmake`:
  - `IGN_INCLUDE_INSTALL_DIR`
  - `IGN_INCLUDE_INSTALL_DIR_POSTFIX`
  - `IGN_INCLUDE_INSTALL_DIR_FULL`
  - `IGN_DATA_INSTALL_DIR_POSTFIX`
  - `IGN_DATA_INSTALL_DIR`
  - `IGN_LIB_INSTALL_DIR`
  - `IGN_BIN_INSTALL_DIR`

- `GzConfigureBuild.cmake`
  - `IGN_CXX_XXX`
  - `IGN_KNOWN_CXX_STANDARDS`
  - CMakeLists.txt search path now prioritizes `gz` over `ignition`

- `GzSetCompilerFlags.cmake`
  - `IGN_ADD_fPIC_TO_LIBRARIES`
  - `IGN_USE_STATIC_RUNTIME` (option and cache variable)

- `GzPython.cmake`
  - `IGN_PYTHON_VERSION`

- Hmhm
  - `REPLACE_IGNITION_INCLUDE_PATH`
  - `NO_IGNITION_PREFIX`


cmake/GzConfigureProject.cmake
```diff
@@ -1,16 +1,16 @@
...
-# IgnConfigureProject
+# GzConfigureProject
 # -------------------
 #
 # gz_configure_project([VERSION_SUFFIX <pre|alpha|beta|etc>])
 #
 # Sets up a Gazebo library project.
 #
-# NO_IGNITION_PREFIX: Optional. Don't use ignition as prefix in
+# NO_GZ_PREFIX: Optional. Don't use Gazebo as prefix in
```
scpeters, line 9: We could go more generic with `NO_PROJECT_PREFIX`

Also, say "Don't use *gz-* as prefix in", since the prefix is `gz` not `Gazebo`

cmake/GzConfigureProject.cmake
```diff
@@ -1,16 +1,16 @@
...
 #     cmake project name.
-# REPLACE_IGNITION_INCLUDE_PATH: Optional. Specify include folder
+# REPLACE_GZ_INCLUDE_PATH: Optional. Specify include folder
```
scpeters, line 11: we could choose a more generic name with `REPLACE_INCLUDE_PATH`

methylDragon: https://github.com/gazebosim/gz-cmake/pull/275/commits/141d2aeb30d48f06b79f23f7cd9bef56d76c6e07

cmake/GzConfigureProject.cmake
methylDragon: https://github.com/gazebosim/gz-cmake/pull/275/commits/141d2aeb30d48f06b79f23f7cd9bef56d76c6e07

methylDragon:
This PR can be squashed

cmake/GzSanitizers.cmake
```diff
@@ -17,7 +17,7 @@
 
 include(CheckCXXSourceCompiles)
 
-set(IGN_SANITIZER ""
+set(GZ_SANITIZER ""
```
chapulina, line 20: This is something we expect users to set, so we should tick-tock it (i.e. also accept `IGN_SANITIZER`, but with warnings).

chapulina (changes requested):
Partial review so far

cmake/GzSanitizers.cmake
methylDragon: https://github.com/gazebosim/gz-cmake/pull/275/commits/84e0509a8895397cf49d00d7ce0fe2534b29512e !

tools/doc_check.sh
```diff
@@ -5,15 +5,15 @@
...
+    cat gz-doxygen.warn
     exit 1
   else
-    echo "Success. No warnings found in ignition-doxygen.warn."
+    echo "Success. No warnings found in gz-doxygen.warn."
     exit 0
   fi
 else
-  echo "The ignition-doxygen.warn file does not exist. Skipping doc check."
+  echo "The gz-doxygen.warn file does not exist. Skipping doc check."
```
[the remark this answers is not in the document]
methylDragon: https://github.com/gazebosim/gz-cmake/pull/275/commits/84e0509a8895397cf49d00d7ce0fe2534b29512e !

Migration.md
```diff
@@ -15,7 +15,10 @@ release will remove the deprecated code.
     **Replacement**: CMake functions and macros starting with `gz_`
 1. **Deprecated**: `ignition` namespaces
     **Replacement**: `gz` namespaces
-
+1. **Deprecated**: `Ign` prefixed CMake files
+    **Replacement**: `Gz` prefixed CMake files
+1. **Deprecated**: `Ign/IGNITION` prefixed CMake variables and options
```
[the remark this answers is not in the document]
methylDragon: https://github.com/gazebosim/gz-cmake/pull/275/commits/84e0509a8895397cf49d00d7ce0fe2534b29512e !

cmake/GzPython.cmake
```diff
@@ -43,3 +43,6 @@ endif()
 if(Python3_EXECUTABLE AND NOT PYTHON_EXECUTABLE)
   set(PYTHON_EXECUTABLE ${Python3_EXECUTABLE})
 endif()
+
+set(IGN_PYTHON_VERSION ${GZ_PYTHON_VERSION} CACHE STRING  # TODO(CH3): Deprecated. Remove on tock.
+  "Deprecated. Use [GZ_PYTHON_VERSION] instead! Specify specific Python version to use ('major.minor' or 'major')")
```
chapulina, line 48: Should this be up top?

cmake/GzPython.cmake
```diff
@@ -13,21 +13,21 @@
...
     "Specify specific Python version to use ('major.minor' or 'major')")
 
   # if not specified otherwise use Python 3
-  if(NOT IGN_PYTHON_VERSION)
-    set(IGN_PYTHON_VERSION "3")
+  if(NOT GZ_PYTHON_VERSION)
+    set(GZ_PYTHON_VERSION "3")
   endif()
 
   find_package(PythonInterp ${IGN_PYTHON_VERSION} QUIET)
```
chapulina, line 33:
```suggestion
  find_package(PythonInterp ${GZ_PYTHON_VERSION} QUIET)
```

cmake/GzSetCompilerFlags.cmake
```diff
@@ -347,14 +348,24 @@ macro(_gz_setup_msvc)
   # framework.
   #
   # In some cases, a user might want to compile with the static runtime. This
-  # should ONLY be done if they do not intend to use the ignition library as
+  # should ONLY be done if they do not intend to use the Gazebo library as
   # part of a plugin-based framework.
-  option(IGN_USE_STATIC_RUNTIME "Use the static runtime (strongly discouraged)" OFF)
+
+  if(IGN_USE_STATIC_RUNTIME)  # TODO(CH3): Deprecated. Remove on tock.
```
chapulina, line 354: Do we still need the `option()` line before checking if it's set?

cmake/GzSetCompilerFlags.cmake
```diff
@@ -347,14 +348,24 @@ macro(_gz_setup_msvc)
...
+    set(IGN_USE_STATIC_RUNTIME ${GZ_USE_STATIC_RUNTIME})
+  endif()
+
   if(BUILD_SHARED_LIBS)
     # Users should not choose the static runtime unless they are compiling a
     # static library, so we completely disable this option if BUILD_SHARED_LIBS
     # is turned on.
-    set(IGN_USE_STATIC_RUNTIME OFF CACHE BOOL "Use the static runtime (strongly discouraged)" FORCE)
+    set(GZ_USE_STATIC_RUNTIME OFF CACHE BOOL "Use the static runtime (strongly discouraged)" FORCE)
+    set(IGN_USE_STATIC_RUNTIME ${GZ_USE_STATIC_RUNTIME}  # TODO(CH3): Deprecated. Remove on tock.
```
chapulina, line 367: If I understand correctly, this is internally setting the variable. So we should only need to set `GZ_` and use that.

chapulina (changes requested):
Finished doing a pass on the code, still need to test locally with other libraries

cmake/GzSetCompilerFlags.cmake
methylDragon: Option will create a cache variable which can then be set via CLI, if I'm not wrong

cmake/GzSetCompilerFlags.cmake
methylDragon: https://github.com/gazebosim/gz-cmake/pull/275/commits/5f4f308a2aa9b85caeaeab20d15cad690e29e545

cmake/GzPython.cmake
methylDragon: [`5f4f308` (#275)](https://github.com/gazebosim/gz-cmake/pull/275/commits/5f4f308a2aa9b85caeaeab20d15cad690e29e545) !

cmake/GzPython.cmake
methylDragon: https://github.com/gazebosim/gz-cmake/pull/275/commits/5f4f308a2aa9b85caeaeab20d15cad690e29e545

PR #277, opened 2022-07-12
ign -> gz Migrate Internal CMake Vars : gz-cmake

methylDragon:
Retarget of: https://github.com/gazebosim/gz-cmake/pull/275

Built on top of: https://github.com/gazebosim/gz-cmake/pull/274

In an effort to keep track of what has already been tick-tocked, I placed the `Deprecated.` comment (in most cases) on the same line so it appears on grep.

Try it: `grep -ri set\(ign`

# BREAKING NOTES
- `IGNITION_DOXYGEN_TAGFILES` is set by DOWNSTREAM libraries and used in this UPSTREAM library... I am just going to hard-tock it and remember to update the downstream usage since it's just for docs...
  - This won't lead to a compile error, but will lead to malformed docs until the downstream libraries are tocked.

The next few ones look like they're breaking (because they change the install path to `gz`), **but local builds work**.
Release repos might need to be updated to have the topline `ignition` -> `gz`.

- Change default include directory from `include/ignition` to `include/gz` (I think in `GzConfigureBuild.cmake` and `GzPackaging.cmake`?)
- `PROJECT_INCLUDE_DIR` stores `ignition/${IGN_DESIGNATION}` -> `gz/...`
  - This would be breaking, but all downstream uses actually bypass this block, so I think it's safe to adjust, I think
  - This crucially changes the `install` directory to `gz` instead of `ignition`, likely necessitating release repository changes, but no other issues if packages are rebuilt accordingly

See: https://github.com/gazebosim/gz-cmake/pull/275/commits/e1c4a355353937c1c11cb2841356e09f348051b2

# Pending??
- [What should be done about `version_info.json.in`](https://github.com/gazebosim/gz-cmake/blob/main/cmake/version_info.json.in#L40) (I've provisionally migrated `ignition` to `gz` here)

# Hard-Tocks
Usually because they're internal or aren't used anywhere else in the entire Gz stack
- GitHub workflow targets

- Erm.. docs only mentions?
  - `IGN_SANITIZERS`
  - `USE_IGN_RECOMMENDED_FLAGS`
- `GzCodeCoverage.cmake`:
  - `IGN_SETUP_TARGET_FOR_COVERAGE`
- Top level gz-cmake CMakeLists.txt
  - `ign_pkgconfig_xxx`
  - `ign_utilities_xxx`
  - `IGN_PC_CONFIG_RELATIVE_PATH_TO_PREFIX`
- `GzPackaging.cmake`:
  - `IGNITION_CMAKE_DIR`
  - `IGNITION_CMAKE_VERSION_MAJOR`
- `GzCreateDocs.cmake`
  - `IGNITION_DOXYGEN_GENTAGFILE`
  - `IGNITION_DOXYGEN_INPUT`
  - `IGNITION_DOXYGEN_IMAGE_PATH`
  - `IGNITION_DOXYGEN_API_MAINPAGE_MD`
  - `IGNITION_DOXYGEN_GENHTML`
  - `IGNITION_DOXYGEN_AUTOGENERATED_DOC`
  - `IGNITION_DOXYGEN_TUTORIALS_MAINPAGE_MD`
  - `IGNITION_DOXYGEN_TUTORIALS_DIR`
  - `IGNITION_DOXYGEN_ADDITIONAL_INPUT_DIRS`
- `GzRonn2Man.cmake`
  - `ign_add_manpage_target`

- `gz-config.cmake.in`
  - `set_and_check(@PKG_NAME@_INCLUDE_DIRS "@PACKAGE_IGN_INCLUDE_INSTALL_DIR_FULL@")` this is the only occurance of that IGN variable. I don't know why

# Tick-Tocks
- Every variable that is set with an IGNITION/IGN prefix, generally in `.cmake` files which are meant to be included (internal references hard-tocked, but the set call itself is ticktocked.)

- `gz-cmake-config.cmake.in`:
  - `IGN_INCLUDE_INSTALL_DIR_FULL`
  - `IGNITION_CMAKE_VERSION_MAJOR`
  - `IGNITION_CMAKE_DOXYGEN_DIR` (Also in top level CMakeLists.txt... it's confusing, I know)
  - `IGNITION_CMAKE_CODECHECK_DIR`
  - `IGNITION_CMAKE_BENCHMARK_DIR`
  - `IGNITION_CMAKE_TOOLS_DIR`

- All these IGN_DESIGNATION variables...
  - `IGN_DESIGNATION`
  - `IGN_DESIGNATION_LOWER`
  - `IGN_DESIGNATION_UPPER`
  - `IGN_DESIGNATION_FIRST_LETTER`
  - `IGN_DESIGNATION_CAP`

- `GzPackaging.cmake`:
  - `IGN_INCLUDE_INSTALL_DIR`
  - `IGN_INCLUDE_INSTALL_DIR_POSTFIX`
  - `IGN_INCLUDE_INSTALL_DIR_FULL`
  - `IGN_DATA_INSTALL_DIR_POSTFIX`
  - `IGN_DATA_INSTALL_DIR`
  - `IGN_LIB_INSTALL_DIR`
  - `IGN_BIN_INSTALL_DIR`

- `GzConfigureBuild.cmake`
  - `IGN_CXX_XXX`
  - `IGN_KNOWN_CXX_STANDARDS`
  - CMakeLists.txt search path now prioritizes `gz` over `ignition`

- `GzSetCompilerFlags.cmake`
  - `IGN_ADD_fPIC_TO_LIBRARIES`
  - `IGN_USE_STATIC_RUNTIME` (option and cache variable)

- `GzPython.cmake`
  - `IGN_PYTHON_VERSION`

- Hmhm
  - `REPLACE_IGNITION_INCLUDE_PATH`
  - `NO_IGNITION_PREFIX`


chapulina (approved):
Works for me locally and I see some of the downstream `ci_matching_branches` also have passing CI. 

While reviewing this I discovered that we have an issue with doxygen on `gz-sim`, which isn't introduced by this PR. I'll either ticket an issue or open a PR with a fix soon.

PR #301, opened 2022-08-10
Build examples using native CMake

mjcarroll:
# 🦟 Bug fix

Fixes (for the most part) Windows examples_build and deduplicates some build steps

## Summary

When testing on the newest conda/Visual Studio 2022/Windows 11 configuration I ran into some issues with the "FAKE_INSTALL" step that we were doing in many packages, which was seen to be invalid.  I was able to reproduce on Linux by trying to use the "ninja" generator for CMake.

After some investigation, I found that FAKE_INSTALL was actually causing the packages to be built twice, which adds significant overhead to the build.

This is an alternative mechanism for building examples as part of the test suite.  Any package that wants to build examples will have to be updated to take advantage of this macro.

This is a pure CMake implementation that doesn't rely on calling out to other scripts/executables.

Downstream PRs:
- [ ] gz-math https://github.com/gazebosim/gz-math/pull/482

## Checklist
- [x] Added tests


cmake/GzUtils.cmake
```diff
@@ -1944,6 +1944,116 @@ macro(gz_build_tests)
...
+#             For example ${CMAKE_CURRENT_SOURCE_DIR}/examples
+#
+# BINARY_DIR: Required. Path to the output binary folder
+#             For example ${CMAKE_CURRENT_BINARY_DIR}/examples
+#
+# COMPONENTS: Optional. List of components built by this project
+#             that are used in the examples.
+#             For example "cli" for gz-utils
+#
+# DEPENCENCIES: Optional.  Gazebo library dependencies for this
```
ahcorde, line 1978:
```suggestion
# DEPENCENCIES: Optional. Gazebo library dependencies for this
```

cmake/GzUtils.cmake
```diff
@@ -1944,6 +1944,116 @@ macro(gz_build_tests)
...
+#
+# BINARY_DIR: Required. Path to the output binary folder
+#             For example ${CMAKE_CURRENT_BINARY_DIR}/examples
+#
+# COMPONENTS: Optional. List of components built by this project
+#             that are used in the examples.
+#             For example "cli" for gz-utils
+#
+# DEPENCENCIES: Optional. Gazebo library dependencies for this
+#               project
```
chapulina, line 1979: It's not clear to me what dependencies need to go here. I see that for `gz-math` you added

```
    DEPENDENCIES 
      gz-cmake${GZ_CMAKE_VER} 
      gz-utils${GZ_UTILS_VER} 
```

Both which aren't explicit dependencies of the examples themselves:

https://github.com/gazebosim/gz-math/blob/29e3f4182037ba8c0d525932bda5c6055fe1bb7c/examples/CMakeLists.txt#L6-L7

Do we need to repeat here all the gz libraries that the parent project depends on, but not that project itself?

cmake/GzUtils.cmake
```diff
@@ -1944,6 +1944,159 @@ macro(gz_build_tests)
...
+#             For example ${CMAKE_CURRENT_SOURCE_DIR}/examples
+#
+# BINARY_DIR: Required. Path to the output binary folder
+#             For example ${CMAKE_CURRENT_BINARY_DIR}/examples
+#
+# COMPONENTS: 
+#     Optional. List of components built by this project that are used in the examples.
+#     For example "cli" for gz-utils
+#
+# DEPENDENCIES:
```
j-rivero, line 1980: Let me see if I understand this right: in `DEPENDENCIES` you can specify gz- libraries that are not being used by the build system of the project but are required to make the examples to build/run? Is this correct? If so, bellow there is a comment:

> we are passing in the resolved dependency directories via -Ddep_DIR=${dep_DIR}

How does this work if a dependency is not in the current buildsystem?

mjcarroll: I have a slight rework to this I can push shortly that drops this gymnastics a bit.  

Note that this is only necessary for gazebo packages, as they may be built in the same workspace, so this special forwarding happens.

Currently, examples aren't allowed to depend on Gazebo packages things that the containing package doesn't.  E.g. gz-gui examples cannot have a dependency on gz-gazebo, as this will introduce circular dependencies.

If an example depends on a non-Gazebo package, then typical `find_package` infrastructure should work.

cmake/GzUtils.cmake
```diff
@@ -1944,6 +1944,161 @@ macro(gz_build_tests)
...
+  # Parse the arguments
+  _gz_cmake_parse_arguments(gz_build_examples "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
+  set(BUILD_EXAMPLES_CMAKE "")
+  set(BUILD_EXAMPLES_INCLUDES "")
+
+  get_target_property(PROJECT_LIBRARY_DEPS ${PROJECT_LIBRARY_TARGET_NAME} INTERFACE_LINK_LIBRARIES)
+
+  # Below here we are generating a series of arguments to the cmake command that is to follow.
+  # Since we may be building in a colcon workspace or otherwise isolated, we are passing
```
azeey, line 1997: It's not clear to me why all of this is needed. I deleted `BUILD_EXAMPLES_CMAKE` and all the related code and the test in gz-math still works. I even added a dependency on `gz-utils2` in the gz-math example just to make sure. It kinda makes sense that it worked because the examples are all self contained and whatever you use to run them, be it colcon or isolated cmake, it would have the same requirements as the package that contains them. If you're able to build the package, I would think you would be able to build the examples without the need to pass resolved dependency directories.

azeey (commented):
I finally got around to reviewing this. Sorry for the long delay. 

My main issue is that we're adding all the resolved dependency directories when it doesn't seem necessary. Maybe something is misconfigured in my local system and there's a real need to do this.

cmake/GzUtils.cmake
mjcarroll: Now it's been a bit, but I seem to recall the issue as follows.  It's not when the examples depend on a package that the core is already depending on (eg math already exists on gz-utils).  The issue arises when examples depend on something that isn't part of the core package, such as an example depending on `gz-utils2::cli`.

mjcarroll: Ah, I remember now, this was extra work so that `make test` and `ctest` work directly from the `build` directory. An example that definitely fails without the dependencies (but passes with `colcon test`)

In gz-common without the dependencies set (these lines removed), it fails:

https://github.com/gazebosim/gz-common/blob/569513a14f703a16a5c7ddd2453c6ce037a03cb9/CMakeLists.txt#L169-L172

So I suppose is if the extra cmake boilerplate is worth it in order to support `make && make test` or `ninja && ninja test` (which is also how we invoke in github actions CI at the moment, iirc)

azeey: Ah, I see the problem now as I'm testing the gz-common PR. The way `colcon test` works is by setting the `CMAKE_PREFIX_PATH` environment variable, so what if we did the same based on the environment variable set during `colcon build`? I tried doing that and it seems to work for both `colcon test` and working directly from the `build` directory.

```cmake

  set(gz_build_examples_CMAKE_PREFIX_PATH $ENV{CMAKE_PREFIX_PATH})
  string(REPLACE ":" ";" gz_build_examples_CMAKE_PREFIX_PATH ${gz_build_examples_CMAKE_PREFIX_PATH})
  list(APPEND gz_build_examples_CMAKE_PREFIX_PATH ${CMAKE_INSTALL_PREFIX})

 # then pass it to `add_test`
  add_test(
    NAME EXAMPLES_Configure_TEST
    COMMAND ${CMAKE_COMMAND} -G${CMAKE_GENERATOR}
                             --no-warn-unused-cli
                             -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
                             -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
                             -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
                             -DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}
                             -DCMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS}
                             -S ${gz_build_examples_SOURCE_DIR}
                             -B ${gz_build_examples_BINARY_DIR}
                             "-DCMAKE_PREFIX_PATH=${gz_build_examples_CMAKE_PREFIX_PATH}"

mjcarroll: Oh, yeah, that should work and be much less code to maintain.  Don't know why I didn't think of that.

azeey (approved):

PR #302, opened 2022-08-12
Drop support for OGRE2 < 2.2

mjcarroll:
# 🦟 Bug fix

## Summary

This is a pretty substantial cleanup of the OGRE2/OGRE-Next logic.

* Allows for OGRE-2.3 packaged by OpenRobotics from packages.osrfoundation.org
* Allows for OGRE-Next from Conda on Windows, Linux, and Mac
* Allow for OGRE-next to be built from source in the same colcon workspace
  * Requires: https://github.com/OGRECave/ogre-next/pull/324
  * Requires: https://github.com/mjcarroll/ogre-next/commit/262471390763eb05088e379291ae9d284b28a05f  

Drops support for < 2.2 and unifies the logic between Linux and Windows.

## Checklist
- [ ] Added tests


mjcarroll:
CC: @darksylinc 

This is going to get us most of the way on deb/conda/source compatibility

cmake/FindGzOGRE2.cmake
```diff
@@ -75,33 +76,18 @@ if (${GzOGRE2_FIND_VERSION_MAJOR})
   endif()
 endif()
 
+function(colcon_do_nothing)
```
j-rivero, line 79: Out of curiosity, how this works when using colcon?

mjcarroll: Colcon is pretty naive when it comes to finding packages.  It does a regex across cmake files looking for `find_package(foo)` to see if there is a dependency on foo.

This forces colcon to understand that ogre-next needs to come before gz-cmake if you are building both in the same workspace.

The alternative is to use `colcon.pkg` or `package.xml`, but I don't think that we have made any progress on that front yet. I can add a todo to remove this when we do something with metadata.

j-rivero: Thanks for the pointer, I think it is fine to add an small comment like this one in the source code by now until we resolve the approaches of package.xml or colcon.pkg.

mjcarroll:
As an alternative @darksylinc has proposed that we may just be able to use  upstream `FindOGRE`, but I haven't tested this on all platforms yet.

j-rivero:
Sorry for being late to review this and now that we have released gz-cmake3 seems a disruptive change to me. Actually the diff does not provide very useful information about what was changed so it is more like a new module from scratch.

Probably at this point there are no more consumers than the Garden family but it would be great to preserve the module as it is if it is not fully compatible with the one in the current code. Several options:

 * **Most compatible change:**  we preserve current module, deprecate it and move these changes to be a new CMake module with a new name. This would require small changes in consumer inside Garden.
 * **Try a transition by default:** we merge these changes but keep a copy of the old module with a different name. Mention it in Changelog/Migration so if it is failing for anyone the change to keep previous behaviour is just a name in `find_package`.

Whatever you prefer @mjcarroll works for me.

j-rivero:
> As an alternative @darksylinc has proposed that we may just be able to use upstream `FindOGRE`, but I haven't tested this on all platforms yet.

Aaah that would be great, indeed.

mjcarroll:
> Probably at this point there are no more consumers than the Garden family but it would be great to preserve the module as it is if it is not fully compatible with the one in the current code. Several options:

That was my thought as well. gz-cmake3 should only be used by the garden collection (unless there are other people out there using it).  I think default behavior should be okay in this case, but I'm happy to preserve the functionality as well.

Since there is no rush, lets hold on this until I have an opportunity to test upstream cmake logic and see if that can replace this.

iche033:
I tested building gz-rendering7 with the changes in this PR and also https://github.com/gazebo-forks/ogre-next/tree/gazebo/v2-3 built from source in a colcon workspace. I saw that gz-rendering was able to find ogre-next 2.3 but had issue with include paths. I get a few of these compile errors:

```sh
/home/iche/code/ogre_ws/install/include/OGRE-Next/Hlms/Unlit/OgreHlmsUnlit.h:32:10: fatal error: OgreHlmsBufferManager.h: No such file or directory
   32 | #include "OgreHlmsBufferManager.h"
```

PR #306, opened 2022-08-18
Add support for windows caching

mjcarroll:
# 🦟 Bug fix

## Summary

Using ccache or buildcache on Windows is not supported when using the `/Zi` flag (ref: https://github.com/ccache/ccache/issues/1040)

The recommended guidance at the moment is to drop to `/Z7` to allow for shared caches to work correctly.

Enabled via: `USE_GZ_RECOMMENDED_MSVC_CACHING` and off by default.

## Checklist
- [ ] Added tests


cmake/GzSetCompilerFlags.cmake
```diff
@@ -342,6 +346,29 @@ macro(_gz_setup_msvc)
     # TODO: What flags should be set for PROFILE and COVERAGE build types?
     #       Is it even possible to generate those build types on Windows?
 
+    # Replace any higher warnings with our selected level
```
j-rivero, line 349: I would prefer not to mess with the `CMAKE_{C/CXX}_` variables if possible (or not to do it more than we are currently doing). Would it work something like `add_compile_options(${MSVC_WARNING_LEVEL})`?

mjcarroll: Yeah, in general, I agree.  The issue here is that MSVC is really dumb about how it parses command line options, and if you have multiple warning switches, they can interact poorly.  Simply appending a new warning level to the end causes warnings to show up on the command line (iirc)

mjcarroll: The warning level bit is actually separate from the caching support here.  It's just to suppress that warning with MSVC

j-rivero: > The warning level bit is actually separate from the caching support here. It's just to suppress that warning with MSVC

The comment was looking to avoid messing with `CMAKE_{C/CXX}_` for the warnings and for the optimization in the lines after. If you have already tested and suffer the multiple compile line options, I'm fine just by adding a comment before the `CMAKE_{C/CXX}_` lines explaining why we are not using`add_compile_options`.

PR #303, opened 2022-08-12
ign -> gz Migrate Ignition Headers : gz-cmake

methylDragon:
See: https://github.com/gazebo-tooling/release-tools/issues/784

PR #309, opened 2022-08-30
2 to 3 20220830

nkoenig:
# ➡️ Forward port\

@methylDragon, can you check this to make sure I did this right?

Port ign-cmake2 to gz-cmake3

Branch comparison: https://github.com/gazebosim/gz-cmake/compare/gz-cmake3...ign-cmake2

**Note to maintainers**: Remember to **Merge** with commit (not squash-merge or rebase)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,17 @@
 ## Gazebo CMake 2.x
 
-### Gazebo CMake 2.14.0 (2022-07-25)
+### Gazebo CMake 2.15.0 (2022-08-29)
+
+1. ign -> gz: add `gz/*` header files
+    * [Pull request #303](https://github.com/gazebosim/gz-cmake/pull/303)
+
+1. Backport `GZ_SANITIZER` variable
+    * [Pull request #294](https://github.com/gazebosim/gz-cmake/pull/294)
+
+1. Update doxygen file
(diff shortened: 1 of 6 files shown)


methylDragon (approved):
Looks about right! For cmake it's a little different since no `config.hh` files are generated. The moved files have their blames intact too.

Rebase merge could work with this also, if you want to squash any non-moving commits, but aside from that :+1:

scpeters:
no, I thought we intentionally didn't add the `gz/utilities/*` header files on this branch? cc @chapulina

scpeters:
> no, I thought we intentionally didn't add the `gz/utilities/*` header files on this branch? cc @chapulina

but then we backported them in #303, hmm...

chapulina:
Yeah we had skipped them on purpose since they're deprecated. If it makes it easier to have the `gz` counterpart so that everything is consistent across all libraries, I think that's ok though

scpeters:
> Yeah we had skipped them on purpose since they're deprecated. If it makes it easier to have the `gz` counterpart so that everything is consistent across all libraries, I think that's ok though

ok, that was my mistake for not catching that in #303, which we've already released. This is fine

scpeters (approved):

PR #95, opened 2020-05-16
Add support for finding several  dependencies without pkg-config

traversaro:
In particular, add a dedicated `ign_manual_search` CMake macro, and then use it in `FindUUID` (fix https://github.com/ignitionrobotics/ign-cmake/issues/64) and in the find scripts for FFMPEG libraries.

PR #315, opened 2022-10-21
FindUUID: Do not wrap LIBRARY_NAMES argument with quotes

traversaro:
# 🦟 Bug fix

Fixes https://github.com/conda-forge/libignition-cmake0-feedstock/issues/32 .

## Summary

Apparently `FindUUID` does not find uuid if pkg-config is not installed in the system, even if https://github.com/gazebosim/gz-cmake/pull/95 should have added support for this. This was due to a bug in https://github.com/gazebosim/gz-cmake/pull/95 that is fixed in this PR: multiple arguments to a parameter should not be passed inside quotes, otherwise there are considered as a single parameter.

For consistency, I also removed the quotes from the other arguments to the same function call.

## Checklist
- [ ] Added tests

### DIFF
--- a/cmake/FindUUID.cmake
+++ b/cmake/FindUUID.cmake
@@ -23,9 +23,9 @@ if (UNIX)
     if(NOT UUID_FOUND)
       include(IgnManualSearch)
       ign_manual_search(UUID
-                        HEADER_NAMES "uuid.h"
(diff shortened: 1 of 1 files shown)


scpeters (approved):

scpeters:
thanks for the fix!

PR #325, opened 2022-11-22
Require Python Interpreter

Bi0T1N:
# 🎉 New feature

This PR refers to https://github.com/gazebosim/sdformat/pull/884 because the changes introduced there require a Python interpreter.

## Checklist
- [ ] Added tests

### DIFF
--- a/cmake/GzPython.cmake
+++ b/cmake/GzPython.cmake
@@ -18,10 +18,10 @@ if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.19")
   set(IGN_PYTHON_VERSION ${GZ_PYTHON_VERSION} CACHE STRING  # TODO(CH3): Deprecated. Remove on tock.
     "Deprecated. Use [GZ_PYTHON_VERSION] instead! Specify specific Python version to use ('major.minor' or 'major')")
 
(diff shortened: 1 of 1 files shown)


scpeters:
this is a breaking change and would require sdformat14 to use gz-cmake4. I think it would be more straightforward to add a `message(FATAL_ERROR "...")` or `message(SEND_ERROR "...")` to the sdformat cmake code if `Python3_EXECUTABLE` is not found

j-rivero:
> this is a breaking change and would require sdformat14 to use gz-cmake4. I think it would be more straightforward to add a `message(FATAL_ERROR "...")` or `message(SEND_ERROR "...")` to the sdformat cmake code if `Python3_EXECUTABLE` is not found

+1 Please note that gz-cmake is not only being consumed by SDFormat but can be used from any other software out there so either we add an API option to make the REQUIRED optional or we leave up to the consumer of this module to check the return value and fail accordingly.

I'm with @scpeters, if you go with the second approach you don't need a new major release of gz-cmake (4.x does not have a timeframe yet).

Thanks for the contribution @Bi0T1N, I'm closing it but feel free to comment or reopen if you have more feedback.

PR #326, opened 2022-11-28
Remove all deprecated ignition API references

j-rivero:
The PR removes all the code that was added for the transition from ignition to gz name.

While I was opening this PR I thought that merging this at this point might complicate the possible forwards/backwards ports of other PRs from other branches :(


cmake/GzBenchmark.cmake
```diff
@@ -59,15 +59,6 @@ function(_gz_add_version_info_target)
   )
 endfunction()
 
-function(ign_add_benchmarks)
-  # TODO(chapulina) Enable warnings after all libraries have migrated.
```
azeey, line 63: Note to self: we need to enable warnings in `gz-cmake3`.

examples/no_gz_prefix/README.md
```diff
@@ -22,9 +22,9 @@ make package_source
 ## Configuring include directory names
 
 This package uses the `REPLACE_INCLUDE_PATH` option in `gz_configure_project`
-to allow a custom include path of `no_gz`, which doesn't start with `ignition/`.
-To confirm, build the package and observe that `AlmostEmpty.cc`
-compiles successfully while including `no_gz/Export.hh`:
+to allow a custom include path of `no_gz`. To confirm, build the package
```
azeey, line 25: Should this be
```suggestion
to allow a custom include path of `no_gz`, which doesn't start with `gz/`. To confirm, build the package 
```

README.md
```diff
@@ -1,7 +1,5 @@
 # Gazebo CMake : CMake Modules for Gazebo Projects
 
-**Maintainer:** grey AT openrobotics.org
```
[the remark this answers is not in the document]
j-rivero: It was yes. I think that this comes from times when grey was the only maintainer of the library and not is a more team effort maintained library. We can list him as the original author if you prefer or list some other maintainers in here for completeness.

cmake/GzBenchmark.cmake
j-rivero: https://github.com/gazebosim/gz-cmake/pull/346

azeey (approved):

PR #337, opened 2023-02-04
Unset package_LIBRARIES cache variable in gz_pkg_check_modules_quiet

iche033:
Signed-off-by: Ian Chen <ichen@openrobotics.org>

# 🦟 Bug fix

## Summary

The value of `package_LIBRARIES` var may persist after calling`gz_pkg_check_modules_quiet`, and this causes an issue with subsequent calls to look for the same package. 

The issue happens for me on macOS (cmake version 3.25.2) when building gz-physics from source, see comment https://github.com/gazebosim/gz-sim/issues/1362#issuecomment-1115780880. `TINYXML2` is searched twice, once when cmake tries to find sdformat and again when finding DART. This results in incorrect library paths being set and linked to by DART:

```
ld: library not found for -lTINYXML2_LIBRARY_/opt/homebrew/Cellar/tinyxml2/9.0.0/lib/libtinyxml2.dylib-NOTFOUND
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

## Checklist
- [ ] Added tests

### DIFF
--- a/cmake/GzPkgConfig.cmake
+++ b/cmake/GzPkgConfig.cmake
@@ -167,6 +167,11 @@ macro(gz_pkg_check_modules_quiet package signature)
       unset(${package}_FOUND CACHE)
       set(${package}_FOUND TRUE)
 
+      # unset cached libraries variable if not empty before calling
+      # gz_pkg_check_modules_quiet to avoid using incorrect libname
+      # in the find_library call
+      unset(${package}_LIBRARIES CACHE)
+
       # For some reason, pkg_check_modules does not provide complete paths to the
       # libraries it returns, even though find_package is conventionally supposed
       # to provide complete library paths. Having only the library name is harmful
--- a/cmake/GzPkgConfig.cmake
+++ b/cmake/GzPkgConfig.cmake
@@ -124,6 +124,10 @@ macro(gz_pkg_check_modules_quiet package signature)
       set(gz_pkg_check_modules_quiet_arg)
     endif()
 
+    # unset LIBRARIES variable in case it contains full paths already
+    # due to previous calls of this function searching for the same package.
+    unset(${package}_LIBRARIES)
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

scpeters:
sounds reasonable. are there any other variables we should unset while we're at it?

scpeters (approved):

iche033:
> sounds reasonable. are there any other variables we should unset while we're at it?

There're also a few variables set by [pkg_check_modules](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html). The ones that are used internally later in our code are:
* `package_FOUND`
* `package_LIBRARIES`
* `package_LIBRARY_DIRS`
* `package_INCLUDE_DIRS`

When I tried to printout these values between two `pkg_check_modules` calls, I see that the `package_LIBRARIES` is the only one whose value came out different after the second call if its previous value is not empty.  I think unsetting other values maybe be fine but I left them untouched to  minimize the chance of breaking things since I'm not very familiar with the implementation details of this function.

scpeters:
> > sounds reasonable. are there any other variables we should unset while we're at it?
> 
> There're also a few variables set by [pkg_check_modules](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html). The ones that are used internally later in our code are:
> 
> * `package_FOUND`
> * `package_LIBRARIES`
> * `package_LIBRARY_DIRS`
> * `package_INCLUDE_DIRS`
> 
> When I tried to printout these values between two `pkg_check_modules` calls, I see that the `package_LIBRARIES` is the only one whose value came out different after the second call if its previous value is not empty. I think unsetting other values maybe be fine but I left them untouched to minimize the chance of breaking things since I'm not very familiar with the implementation details of this function.

sounds good! I approve

PR #342, opened 2023-04-05
Fix finding ogre 2.3 installed from source

iche033:
# 🦟 Bug fix

## Summary

When building gz with ogre-next 2.3 from source, cmake has trouble finding ogre-next because the `OGRE2_LIBRARY_DIRS` is empty and `OGRE2_RESOURCE_PATH` points to the incorrect dir. 
* `OGRE2_LIBRARY_DIRS`: Turns out that variable is set by a `STRING` `REPLACE` call that never gets called because the if check before string replace seems incorrect - I updated to make sure the input string exists before doing the replace call (originally the logic checks that the input string does not exist).
* `OGRE2_RESOURCE_PATH`: when built from source, the ogre plugins are installed to a dir called `OGRE-Next` instead of `OGRE`

Tested with ogre-next 2.3 installed from https://github.com/gazebo-forks/ogre-next/tree/gazebo/v2-3

## Checklist
- [ ] Added tests


mjcarroll:
I think I addressed this in https://github.com/gazebosim/gz-cmake/pull/302, though it is a larger change

iche033:
> I think I addressed this in #302, though it is a larger change

ah I see, I'll give that a try

Issue #9, opened 2017-12-06
FindIgnProtobuf could be more specific when libprotoc-dev is missing


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-cmake/issues/9)) by Shane Loretz (Bitbucket: [Shane Loretz](https://bitbucket.org/%7B656e3311-aad9-45a1-aaf7-b0ee0e84b287%7D/), GitHub: [sloretz](https://github.com/sloretz)).**

----------------------------------------
While cleaning up a docker image based on 16.04 I had installed `libprotobuf-dev` and `protobuf-compiler` but accidentally removed `libprotoc-dev`. 

The output at generate time is

```
-- Found Protobuf: /usr/lib/x86_64-linux-gnu/libprotobuf.so (Required is at least version "2.3.0") 
-- Looking for Protobuf - found
```

but the output at build time is 

```
-- Build files have been written to: /workspace/ws_ign/build_isolated/ign_msgs
==> '. /workspace/ws_ign/build_isolated/ign_msgs/cmake__build.sh && /usr/bin/make -j8 -l8' in '/workspace/ws_ign/build_isolated/ign_msgs'
src/CMakeFiles/ign_msgs_gen.dir/build.make:120: *** target pattern contains no '%'.  Stop.
CMakeFiles/Makefile2:1186: recipe for target 'src/CMakeFiles/ign_msgs_gen.dir/all' failed
```

Line 120 in build.make is `src/ign_msgs_gen: protobuf::libprotoc-NOTFOUND`

I'd like the the generate step to error with "libprotoc not found". How could that be accomplished? Does it warrant it's own find module? Or would it fit better as a `COMPONENT` of `FindIgnProtobuf`?

Ryanf55:
I took a look at this, can we just remove the FindGzProtobuf.cmake entirely? 

Since CMake 3.9, there are numerous improvements to the [FindProtobuf.cmake](https://cmake.org/cmake/help/latest/module/FindProtobuf.html). I'll submit a PR to try to use that instead.

mjcarroll:
> I took a look at this, can we just remove the FindGzProtobuf.cmake entirely?

I think that it probably makes sense.  I noticed recently that our implementation misses some of those improvements.  This would be a good opportunity to do that.

Ryanf55:
Sounds good. 

[...]

PR #363, opened 2023-07-07
Support find-protobuf using standard CMake

Ryanf55:
## Summary

This removes the custom FindGzProtobuf since one is already supplied by CMake and we have new enough CMake available to use it. Relates to https://github.com/gazebosim/gz-cmake/issues/9#issuecomment-1625729540

A few more PR's are going to be pushed up too.

## Checklist
- [ ] Added tests


traversaro:
Unfortunatly at the moment the `FindProtobuf.cmake` shipped with CMake does not work correctly with protobuf >= 23.2, see https://github.com/gazebosim/gz-msgs/pull/346 and https://gitlab.kitware.com/cmake/cmake/-/issues/24321 . If `FindProtobuf.cmake` could be used as it is, I guess we can't just remove the module from an already publish released of gz-cmake .

scpeters:
> Unfortunatly at the moment the `FindProtobuf.cmake` shipped with CMake does not work correctly with protobuf >= 23.2, see [gazebosim/gz-msgs#346](https://github.com/gazebosim/gz-msgs/pull/346) and https://gitlab.kitware.com/cmake/cmake/-/issues/24321 . If `FindProtobuf.cmake` could be used as it is, I guess we can't just remove the module from an already publish released of gz-cmake .

I think we could consider deprecating it on the `main` branch, but I think the first step would be to change the packages (like gz-msgs*) that use this find module over to the `find_package` call you suggest instead.

Ryanf55:
> > Unfortunatly at the moment the `FindProtobuf.cmake` shipped with CMake does not work correctly with protobuf >= 23.2, see [gazebosim/gz-msgs#346](https://github.com/gazebosim/gz-msgs/pull/346) and https://gitlab.kitware.com/cmake/cmake/-/issues/24321 . If `FindProtobuf.cmake` could be used as it is, I guess we can't just remove the module from an already publish released of gz-cmake .
> 
> I think we could consider deprecating it on the `main` branch, but I think the first step would be to change the packages (like gz-msgs*) that use this find module over to the `find_package` call you suggest instead.

Sounds good. We can start with the other PR's first.

traversaro:
The problem that I see is that if you use find_package(Protobuf MODULE) we break the compilation against protobuf >= 23, while if you use find_package(Protobuf CONFIG) we break the compilation of autotools-compiled Protobuf (that includes the Protobuf Debian packages in almost all Ubuntu distros).

Ryanf55:
> The problem that I see is that if you use find_package(Protobuf MODULE) we break the compilation against protobuf >= 23, while if you use find_package(Protobuf CONFIG) we break the compilation of autotools-compiled Protobuf (that includes the Protobuf Debian packages in almost all Ubuntu distros).

Hmm, yea. Thanks, it seems you've spent a bunch of time on this. 

 It's unfortunate some errors are during linking, the major distros are building with autotools, the protobuf maintainers don't seem bothered they broke CMake support on major platforms, even if we patch CMake's find module today, it will only help if you have newer than CMake 3.28.
 
 Things to figure out:
 * Which major distros in gazebo are affected? 
   * For me on Ubuntu 23, I am unaffected, but it's a moot point because protobuf isn't claiming support on non-LTS Ubuntu releases.
 * Does gazebo have to support source compilation on all combinations of host OS + gazebo version + package-installed protobuf or source-compiled-protobuf, or are there combinations that can be ignored? 
 
 Ideas:
 * Request package maintainers to switch to CMake instead of autotools, mainly Ubuntu 24
 * Request CMake backport a bugfix for version handling in their find module to the CMake versions released by major distros
 * Split the difference and try some combination of `find_package(Protobuf MODULE)` and `find_package(Protobuf CONFIG)` with perhaps some custom version handling detection for gazebo until either of the above are done

traversaro:
> Split the difference and try some combination of `find_package(Protobuf MODULE)` and `find_package(Protobuf CONFIG)` with perhaps some custom version handling detection for gazebo until either of the above are done

The problem is that at that point if you have boilerplate you may want to avoid duplicating it across gz-* repos, and that is thea reason why `FindGzProtobuf.cmake` exists in the first place.

Ryanf55:
> > Split the difference and try some combination of `find_package(Protobuf MODULE)` and `find_package(Protobuf CONFIG)` with perhaps some custom version handling detection for gazebo until either of the above are done
> 
> The problem is that at that point if you have boilerplate you may want to avoid duplicating it across gz-* repos, and that is thea reason why `FindGzProtobuf.cmake` exists in the first place.

Would you recommend trying to modify FindGzProtobuf then to do the following? 
1. Try to find protobuf it in CONFIG mode with protobuf's distributed config file, which will only happen if someone built and installed it from source with CMake, since most package managers today supply the autotools version
2. Try to find protobuf in MODULE mode with CMake's supplied version?
3. Fall back to custom logic that hopefully can be deprecated once upstream issues are resolved

traversaro:
> 
> Would you recommend trying to modify FindGzProtobuf then to do the following?
> 
>     1. Try to find protobuf it in CONFIG mode with protobuf's distributed config file, which will only happen if someone built and installed it from source with CMake, since most package managers today supply the autotools version
> 
>     2. Try to find protobuf in MODULE mode with CMake's supplied version?
> 
>     3. Fall back to custom logic that hopefully can be deprecated once upstream issues are resolved



> > > Split the difference and try some combination of `find_package(Protobuf MODULE)` and `find_package(Protobuf CONFIG)` with perhaps some custom version handling detection for gazebo until either of the above are done
> > 
> > 
> > The problem is that at that point if you have boilerplate you may want to avoid duplicating it across gz-* repos, and that is thea reason why `FindGzProtobuf.cmake` exists in the first place.
> 
> Would you recommend trying to modify FindGzProtobuf then to do the following?
> 
>     1. Try to find protobuf it in CONFIG mode with protobuf's distributed config file, which will only happen if someone built and installed it from source with CMake, since most package managers today supply the autotools version
> 
>     2. Try to find protobuf in MODULE mode with CMake's supplied version?
> 
>     3. Fall back to custom logic that hopefully can be deprecated once upstream issues are resolved

If I am not wrong, `FindGzProtobuf` already does 1. (https://github.com/gazebosim/gz-cmake/blob/gz-cmake3_3.2.2/cmake/FindGzProtobuf.cmake#L29) and 2. (https://github.com/gazebosim/gz-cmake/blob/gz-cmake3_3.2.2/cmake/FindGzProtobuf.cmake#L34), while there is no custom (find) logic).

Ryanf55:
Closing for now to reduce the noise. It can be revisited when CMake+Protobuf work out of the box nicely again.

PR #377, opened 2023-08-08
Disable building examples by default

mjcarroll:
# 🦟 Bug fix

Introduced by: https://github.com/gazebosim/gz-cmake/pull/301
Fixes https://github.com/gazebosim/gz-math/issues/539

## Summary

While the new cmake-native mechanism for building examples allows it to work on Windows, it also regresses on non-colcon builds.

When colcon builds run, each package is called with `cmake && make && make install` during the `colcon build` phase, and then `make test` is called during the `colcon test` phase.  This guarantees that the package is installed _somewhere_ before attempting to run the tests, which is what the examples need.

In builds like the deb-builder, this sequence is instead `make && make test` to verify that the package has been correctly built before creating the deb.

This PR makes building examples off by default, and can be enabled by adding `-DBUILD_EXAMPLES:bool=True` to the cmake args for local and CI builds.

## Checklist
- [ ] Added tests

### DIFF
--- a/cmake/GzBuildExamples.cmake
+++ b/cmake/GzBuildExamples.cmake
@@ -40,46 +40,52 @@
 #             For example ${CMAKE_CURRENT_BINARY_DIR}/examples
 #
 macro(gz_build_examples)
-  #------------------------------------
-  # Define the expected arguments
-  set(options)
-  set(oneValueArgs SOURCE_DIR BINARY_DIR)
+  option(BUILD_EXAMPLES "Build examples" OFF)
 
-  #------------------------------------
-  # Parse the arguments
-  _gz_cmake_parse_arguments(gz_build_examples "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+  if(NOT ${BUILD_EXAMPLES})
+    message(STATUS "Building examples disabled")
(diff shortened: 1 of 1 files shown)


mjcarroll:
As part of this, we probably want to also update documentation to reflect that a make install is needed for examples, as well as enabling this in CI.

iche033:
add a note to migration guide?

azeey (approved):

cmake/GzBuildExamples.cmake
```diff
@@ -40,46 +40,52 @@
 #             For example ${CMAKE_CURRENT_BINARY_DIR}/examples
 #
 macro(gz_build_examples)
-  #------------------------------------
-  # Define the expected arguments
-  set(options)
-  set(oneValueArgs SOURCE_DIR BINARY_DIR)
+  option(BUILD_EXAMPLES "Build examples" OFF)
```
j-rivero, line 43:
```suggestion
  option(BUILD_EXAMPLES "Build examples (requires a system installation of gz-cmake)" OFF)
```

j-rivero (approved):

mjcarroll:
Updated migration guide

PR #380, opened 2023-08-24
Only link against DL in the case that it is needed

mjcarroll:
https://github.com/gazebosim/gz-cmake/pull/334 added the ability to do binary relocatability.  

When enabled for shared libraries, it requires the additional dependence on `dl` for the `dladdr` symbol.  This was being unconditionally linked, even in the case that relocation was turned off.

This PR scopes the `dl` dependency to only when it is required.


azeey (approved):
LGTM. I haven't tested locally though.

traversaro:
Thanks for catching this!

PR #382, opened 2023-08-25
Move DL fatal message

mjcarroll:
Moved the logic, but not the error.  I think the DL_target was still defined in my cmake cache so I missed it.

### DIFF
--- a/cmake/GzRelocatableBinaries.cmake
+++ b/cmake/GzRelocatableBinaries.cmake
@@ -74,17 +74,10 @@ macro(gz_add_get_install_prefix_impl)
       "gz_add_get_install_prefix_impl: missing parameter OVERRIDE_INSTALL_PREFIX_ENV_VARIABLE")
   endif()
 
-
   if(NOT TARGET ${PROJECT_LIBRARY_TARGET_NAME})
     message(FATAL_ERROR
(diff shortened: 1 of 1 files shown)


cmake/GzRelocatableBinaries.cmake
```diff
@@ -74,19 +74,12 @@ macro(gz_add_get_install_prefix_impl)
...
   endif()
 
-
   if(NOT TARGET ${PROJECT_LIBRARY_TARGET_NAME})
     message(FATAL_ERROR
       "Target ${PROJECT_LIBRARY_TARGET_NAME} required by gz_add_get_install_prefix_impl\n"
       "does not exist.")
   endif()
 
-  if(NOT TARGET ${DL_TARGET})
```
azeey, line 84: Since there's another code block that's conditioned on `GZ_ENABLE_RELOCATABLE_INSTALL` on line 101(before PR),  should this stay here with an additional condition on `GZ_ENABLE_RELOCATABLE_INSTALL`?

mjcarroll: We don't define GZ_ENABLE_RELOCATABLE_INSTALL until after where the code was before the PR.  I moved it to the chunk in line 101.

azeey (approved):

PR #395, opened 2023-11-10
Update github action workflows

azeey:
* Use on `push` only on stable branches to avoid duplicate runs (see [syntax](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet))
* Update project automation

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,12 @@
 name: Ubuntu CI
 
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+    branches:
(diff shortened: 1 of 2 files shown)


scpeters:
I think the project board automation isn't working

azeey:
> I think the project board automation isn't working

It won't work until we merge this. It uses the `pull_request_target` event in order to access `secrets.TRIAGE_TOKEN`, but that also means it will use the base branch of the PR, not the branch that has the changes for the PR.

mjcarroll (approved):

PR #400, opened 2023-12-22
Add macros for generating pybind11 bindings

### LINKED ISSUE #215: Add CMake macro for generating pybind11 bindings
## Desired behavior
Right now there is some code duplication in [ign-gazebo](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo6/CMakeLists.txt#L179) and [ign-math](https://github.com/ignitionrobotics/ign-math/blob/ign-math6/CMakeLists.txt#L72) for finding pybind11. I assume the goal is to provide Python interfaces also for other packages, thus it makes sense to have a common macro for it.

## Implementation suggestion
Since [`PythonLibs`](https://cmake.org/cmake/help/latest/module/FindPythonLibs.html) is deprecated, the snippet should also use [FindPython3](https://cmake.org/cmake/help/latest/module/FindPython3.html) which will be provided with PR #212.
It might be useful to extend the new `IgnPython` to provide several arguments as input in order to reuse the existing code but this might require some additional effort than creating an independent `IgnPythonBindings` macro.

Bi0T1N:
# 🎉 New feature

Closes #215

## Summary
This adds CMake macros to remove code duplication for creating pybind11 bindings.

The only problem that occurs is that it destroys the used structure in `CMakeLists.txt` (e.g. for gz-math):
```
#============================================================================
# Set project-specific options
#============================================================================

option(SKIP_SWIG
      "Skip generating ruby bindings via Swig"
      OFF)


include(GzPython) <-- here we don't only set project-specific options - we search for Python
gz_add_pybind_project_settings()
```
The other two macros are fine:
```
########################################
# Python bindings
gz_search_for_pybind()
```
`src/CMakeLists.txt`:
```
# Bindings subdirectories
gz_add_pybind_directory(python_pybind11)
```

## Test it
1. Compile packages up to gz-math7 to `/ws/install`
2. In gz-cmake project, run `mkdir build && cd build && cmake .. -DCMAKE_INSTALL_PREFIX=/ws/install/`
3. Run `make -j && make install`
4. Open a new terminal and change directory to gz-math
5. Run `source /ws/install/setup.bash`, change `CMakeLists.txt` to use `gz-cmake4` and the new macros
6. Run `mkdir build && cd build && cmake .. && make -j`

## Checklist
- [ ] Added tests


azeey (approved):

azeey:
LGTM! but I'll hold off merging so @j-rivero can take a look since he commented on #215.

cmake/GzPython.cmake
```diff
@@ -24,3 +24,101 @@ find_package(Python3 ${GZ_PYTHON_VERSION} QUIET)
...
+# Variables defined by this macro:
+#
+#  Python3_*            Python result variables from FindPython3 module
+#  pybind11_*           pybind11 result variables from pybind11 module
+#
+macro(gz_search_for_pybind)
+  if (SKIP_PYBIND11)
+    message(STATUS "SKIP_PYBIND11 set - disabling python bindings")
+  else()
+    find_package(Python3 ${GZ_PYTHON_VERSION} QUIET COMPONENTS Interpreter Development)
```
azeey, line 84: https://github.com/gazebosim/gz-math/pull/588 updated the code in gz-math to use `OPTIONAL_COMPONENTS` here.
https://github.com/gazebosim/gz-math/blob/ef2211c6292929ea6a8471815cf83315c1b891cb/CMakeLists.txt#L102-L104

Mind updating this to match?

cc @scpeters

cmake/GzPython.cmake
```diff
@@ -24,3 +24,101 @@ find_package(Python3 ${GZ_PYTHON_VERSION} QUIET)
...
+#
+#     gz_add_pybind_project_settings()
+#
+# Variables defined by this macro:
+#
+#  SKIP_PYBIND11                                Skip generating of Python bindings if OFF
+#  USE_SYSTEM_PATHS_FOR_PYTHON_INSTALLATION     Install modules in default system path if ON
+#  USE_DIST_PACKAGES_FOR_PYTHON                 Use dist-package over site-package for module installation if ON
+#
+macro(gz_add_pybind_project_settings)
```
scpeters, line 43: apologies for not reviewing this sooner

I actually deprecated `GzPython` in https://github.com/gazebosim/gz-cmake/pull/431; I think these new macros should be in a new file like `GzPybind.cmake`

cmake/GzPython.cmake
scpeters: I think I haven't been completely consistent with using `COMPONENTS` or `OPTIONAL_COMPONENTS`

scpeters: ok, in gz-math8 and sdformat15, I used made finding the Python3 Interpreter `REQUIRED` and used `OPTIONAL_COMPONENTS` to find Development, while we didn't apply this logic in gz-transport or gz-sim, probably because I didn't realize the inconsistency

* https://github.com/gazebosim/gz-math/blob/main/CMakeLists.txt#L102-L105
* https://github.com/gazebosim/sdformat/blob/sdf14/CMakeLists.txt#L87-L90
* https://github.com/gazebosim/gz-transport/blob/main/CMakeLists.txt#L61
* https://github.com/gazebosim/gz-sim/blob/main/CMakeLists.txt#L209

I think I do recommend the approach used in gz-math and sdformat, but I also recognize the lack of flexibility in finding packages from a macro. I would suggest calling `find_package(Python3 ...)` from within the macro if `NOT Python3_Development_FOUND`, which would allow a package to use their own call to `find_package(Python3)` but fall back to a good default if it hasn't been searched for yet

azeey:
To focus our efforts on the Ionic release, I'll go ahead and remove the `beta` label from this. I don't think this is a breaking change, so we can merge it after Ionic.

cmake/GzPython.cmake
Bi0T1N: Implemented as suggested in `GzPybind.cmake` with `OPTIONAL_COMPONENTS` as Python might be used for other tasks where the `Development` component is not required.

azeey:
@j-rivero to take a look.

caguero:
@j-rivero , are we good to go here?

PR #413, opened 2024-02-23
Add package.xml

azeey:
# 🎉 New feature

## Summary

Add `package.xml` based on dependencies found in `.github/ci/packages`. 

This is to test out using vendor packages to provide Gazebo packages to ROS users (see https://github.com/gazebo-tooling/release-tools/issues/1117). 

## Checklist
- [ ] Added tests

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -14,6 +14,8 @@
   <buildtool_depend>cmake</buildtool_depend>
   <buildtool_depend>pkg-config</buildtool_depend>
 
+  <test_depend>cmake</test_depend>
+
   <export>
     <build_type>cmake</build_type>
   </export>
--- a/package.xml
+++ b/package.xml
@@ -2,10 +2,10 @@
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="2">
   <name>gz-cmake3</name>
-  <version>3.4.1</version>
+  <version>3.5.0</version>
   <description>Gazebo CMake : CMake Modules for Gazebo Projects</description>
 
-  <maintainer email="grey@openrobotics.org">Michael X Grey</maintainer>
+  <maintainer email="scpeters@openrobotics.org">Steve Peters</maintainer>
 
   <license>Apache License 2.0</license>
 
--- a/package.xml
+++ b/package.xml
@@ -2,7 +2,7 @@
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="2">
   <name>gz-cmake3</name>
-  <version>3.5.0</version>
+  <version>3.5.1</version>
   <description>Gazebo CMake : CMake Modules for Gazebo Projects</description>
 
   <maintainer email="scpeters@openrobotics.org">Steve Peters</maintainer>
--- a/package.xml
+++ b/package.xml
@@ -2,7 +2,7 @@
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="2">
   <name>gz-cmake3</name>
-  <version>3.5.1</version>
+  <version>3.5.2</version>
   <description>Gazebo CMake : CMake Modules for Gazebo Projects</description>
 
   <maintainer email="scpeters@openrobotics.org">Steve Peters</maintainer>
--- a/package.xml
+++ b/package.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>gz-cmake3</name>
+  <version>3.4.1</version>
+  <description>Gazebo CMake : CMake Modules for Gazebo Projects</description>
+
+  <maintainer email="grey@openrobotics.org">Michael X Grey</maintainer>
+
+  <license>Apache License 2.0</license>
+
+  <url type="website">https://github.com/gazebosim/gz-cmake</url>
+
+  <buildtool_depend>cmake</buildtool_depend>
+  <buildtool_depend>pkg-config</buildtool_depend>
+
+  <export>
+    <build_type>cmake</build_type>
+  </export>
+</package>
--- a/.github/workflows/package_xml.yml
+++ b/.github/workflows/package_xml.yml
@@ -8,4 +8,4 @@ jobs:
     runs-on: ubuntu-latest
     name: Validate package.xml
     steps:
-      - uses: gazebo-tooling/action-gz-ci/validate_package_xml@add_validate_package_xml
+      - uses: gazebo-tooling/action-gz-ci/validate_package_xml@jammy
--- a/.github/workflows/package_xml.yml
+++ b/.github/workflows/package_xml.yml
@@ -1,0 +1,11 @@
+name: Validate package.xml
+
+on:
+  pull_request:
+
+jobs:
+  package-xml:
+    runs-on: ubuntu-latest
+    name: Validate package.xml
+    steps:
+      - uses: gazebo-tooling/action-gz-ci/validate_package_xml@add_validate_package_xml


package.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>gz-cmake3</name>
+  <version>3.4.1</version>
+  <description>Gazebo CMake : CMake Modules for Gazebo Projects</description>
+
+  <maintainer email="grey@openrobotics.org">Michael X Grey</maintainer>
```
sloretz, line 8: Is this email still correct?

azeey: Well, this reminded me to talk to Grey and we've decided to change the maintainer to Steve. Updated in [0a7a257](https://github.com/gazebosim/gz-cmake/pull/413/commits/0a7a257b43c760ad4356efa95114496e0d4179b6)

package.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>gz-cmake3</name>
+  <version>3.5.0</version>
```
scpeters, line 5: I think we should have a plan for how to keep this version tag in sync with the version in the CMakeLists.txt.

Some options:

1. Manually update version numbers in both places and add a CI check to verify consistency (presumably with a shared github action that we will use in other packages)
2. Make this package.xml version string the single source of truth and add cmake logic to parse the package.xml file to get it (presumably using the existing code that ROS / ament is already using)

I won't block this pull request on that change, but I'd like to at least have an issue open where we continue this conversation

scpeters (commented):
are package.xml files typically installed?

package.xml
scpeters: I think option 1 would be easier to implement

scpeters: > add cmake logic to parse the package.xml file to get it (presumably using the existing code that ROS / ament is already using)

I believe the following `ament_cmake` macro is used to invoke the `ament_package` python scripts:

* https://github.com/ament/ament_cmake/blob/9743c9cd3e5aef9566acefea6e6e3724e0e5657e/ament_cmake_core/cmake/core/ament_package_xml.cmake
* https://github.com/ament/ament_package

azeey:
> > are package.xml files typically installed?
> 
> for ROS packages, it looks like the `package.xml` files are installed to `{install_prefix}/share/{package_name}/package.xml`

For the purposes of creating the ROS vendor packages, I don't think we need to install `package.xml`. In general, I was hoping to avoid any changes to our CMake code to support the `package.xml` file and instead treat it as a convenient place  to put our dependencies (so we can use `rosdep`) and also as the source of truth for generating/updating vendor packages with automated tools.

scpeters:
> For the purposes of creating the ROS vendor packages, I don't think we need to install `package.xml`. In general, I was hoping to avoid any changes to our CMake code to support the `package.xml` file and instead treat it as a convenient place to put our dependencies (so we can use `rosdep`) and also as the source of truth for generating/updating vendor packages with automated tools.

sounds good

scpeters (approved):

mjcarroll (approved):

package.xml
scpeters: I just hacked up a crude github workflow check for consistent versions in the CMakeLists.txt and package.xml file in branch https://github.com/gazebosim/gz-cmake/compare/package_xml...package_xml_ci_test

@j-rivero where would you like to see this CI configuration stored?

azeey: I should have mentioned I've added a workflow in some of the other package.xml PRs. See https://github.com/gazebosim/gz-sim/actions/runs/8334724361

package.xml
```diff
@@ -0,0 +1,20 @@
...
+  <description>Gazebo CMake : CMake Modules for Gazebo Projects</description>
+
+  <maintainer email="scpeters@openrobotics.org">Steve Peters</maintainer>
+
+  <license>Apache License 2.0</license>
+
+  <url type="website">https://github.com/gazebosim/gz-cmake</url>
+
+  <buildtool_depend>cmake</buildtool_depend>
```
j-rivero, line 14:
```suggestion
  <depend>cmake</depend>
```
Probably not a big deal but we can declare the `cmake` dependency as build-time, exec-time and build-export-depend dependency since we probably expect `cmake` to be available in all these scenarios.

pkg-config could fit into exec-time but not in all scenarios probably. Good as it is for me.

azeey: Is the question whether we expect `cmake` to be available or that we only need it at build time? I was under the impression that we want to narrow down the scope as much as possible.

scpeters: I think we don't use `cmake` when executing gazebo through any of the `gz` commands, but we do invoke `cmake` in some of our tests; is that a valid reason for the difference between `buildtool_depend` and `depend`?

azeey: I think `<buildtool_depend>cmake</buildtool_depend>` is the right thing here. I have looked at a few examples:

- https://github.com/eclipse-iceoryx/iceoryx/blob/e274957ac619cd9881687e688e3f8d1c28ff115c/iceoryx_binding_c/package.xml#L13
- https://github.com/eProsima/Fast-DDS/blob/ee0637d7a3302a1c69377a0282da7b0f7096292a/package.xml#L25
- https://github.com/ros-industrial/stomp/blob/f9e9d78f741f0779b65333d4f54d68ade740520b/package.xml#L14

It is also what's recommended in https://ros.org/reps/rep-0136.html

> I think we don't use cmake when executing gazebo through any of the gz commands, but we do invoke cmake in some of our tests; is that a valid reason for the difference between buildtool_depend and depend?

Ah, perhaps we'll need `<test_depend>cmake</test_depend>` to cover that.

package.xml
azeey: I've added an Action in https://github.com/gazebo-tooling/action-gz-ci/pull/70 and updated this PR to use it.

azeey:
https://github.com/gazebo-tooling/action-gz-ci/pull/70 adds validation of the package.xml using `ament_xmllint` and  it seems like dashes (`-`) are not allowed in `<name>` tags 😱. Should we switch to underscores for just the package.xml?

scpeters:
> [gazebo-tooling/action-gz-ci#70](https://github.com/gazebo-tooling/action-gz-ci/pull/70) adds validation of the package.xml using `ament_xmllint` and it seems like dashes (`-`) are not allowed in `<name>` tags 😱. Should we switch to underscores for just the package.xml?

from https://ros.org/reps/rep-0149.html#name

> The package name must start with a letter and contain only lowercase alphabetic, numeric or underscore characters

but also

> The following recommended exemptions apply, which are optional for implementations:
> * Dashes may be permitted in package names. This is to support maintaining a consistent dependency name when transitioning back and forth between a system dependency and in-workspace package, since many rosdep keys contain dashes (inherited from the Debian/Ubuntu name).

so maybe ignore that specific linter complaint?

azeey:
> > [gazebo-tooling/action-gz-ci#70](https://github.com/gazebo-tooling/action-gz-ci/pull/70) adds validation of the package.xml using `ament_xmllint` and it seems like dashes (`-`) are not allowed in `<name>` tags 😱. Should we switch to underscores for just the package.xml?
> 
> from https://ros.org/reps/rep-0149.html#name
> 
> > The package name must start with a letter and contain only lowercase alphabetic, numeric or underscore characters
> 
> but also
> 
> > The following recommended exemptions apply, which are optional for implementations:
> > 
> > * Dashes may be permitted in package names. This is to support maintaining a consistent dependency name when transitioning back and forth between a system dependency and in-workspace package, since many rosdep keys contain dashes (inherited from the Debian/Ubuntu name).
> 
> so maybe ignore that specific linter complaint?

This was discussed at today's ROS 2 meeting. The conclusion was that the REP is the authority here. The plan is to relax the naming requirement in the .xsd and potentially provide an opt-in linter that warns against use of dashes. See https://github.com/ros-infrastructure/rep/pull/400.

In the meantime, we can comment out the `ament_xml_lint` from our CI and merge this.

package.xml
```diff
@@ -0,0 +1,20 @@
...
+  <buildtool_depend>pkg-config</buildtool_depend>
```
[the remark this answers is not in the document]
azeey: Added in https://github.com/gazebosim/gz-cmake/pull/413/commits/0f91fadf5dde649b332d601e6fc18895d2ef2a7a

package.xml
azeey: I've added `<test_depend>cmake</test_depend>` in https://github.com/gazebosim/gz-cmake/pull/413/commits/0f91fadf5dde649b332d601e6fc18895d2ef2a7a

scpeters (approved):

PR #425, opened 2024-05-02
Fix installation of `Ign*.cmake` modules on newer versions of CMake

azeey:
# 🦟 Bug fix

## Summary
Previously, we expected `install(CODE ...)`,  to run before `install(FILES ...)` since that is the order they appear in the CMakeLists.txt file. `install(CODE ...)` was responsible for creating a copy or symlink of `Gz*.cmake` files to `Ign*.cmake` 
 and putting them in a temporary location.   `install(FILES ...)` would then copy those files to the final install destination. However, it appears that on CMake 3.29.2 (tested on windows), we can no longer rely on the order they appear in the code. From my local testing, `install(CODE ...)` ran after `install(FILES ...)`.

This PR uses `add_custom_command` to generate the `Ign*.cmake` files which avoids any reliance on the order of the `install` calls.

This was caught during the ROS 2 tutorial party (see https://github.com/osrf/ros2_test_cases/issues/1526)

cc @clalancette 

## Checklist
- [ ] Added tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,19 +182,23 @@ set(tick_tocked_cmake_files
 foreach(cmake_file ${tick_tocked_cmake_files})
   string(REGEX REPLACE "^Gz" "Ign" ign_cmake_file ${cmake_file})
   file(MAKE_DIRECTORY "${PROJECT_BINARY_DIR}/cmake")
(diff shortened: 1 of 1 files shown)


clalancette (approved):
This looks like a good fix to me.  Thanks @azeey !

scpeters (approved):

PR #435, opened 2024-07-20
Accept arbitrary capitalization for coverage build type

azeey:
# 🦟 Bug fix

## Summary
When running `cmake -DCMAKE_BUILD_TYPE=coverage` (which most of our CI does), gz-cmake generates a warning
```
Code coverage results with an optimized (non-Debug) build may be misleading
```

This is only because we are specifically looking for `Coverage`, with uppercase `C`. This patch makes it possible to pass any capitalization of the word.

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #437, opened 2024-07-20
verify warning with -DCMAKE_BUILD_TYPE=coverage

scpeters:
confirming existence of bug to be fixed by #435


scpeters:
hm, this doesn't generate a warning, I guess the example coverage builds are not tested in CI because `lcov` is not installed

scpeters:
> hm, this doesn't generate a warning, I guess the example coverage builds are not tested in CI because `lcov` is not installed

reported in https://github.com/gazebosim/gz-cmake/issues/440

PR #472, opened 2025-02-11
Tentative fix for `FindOptiX.cmake`

roncapat:
Lost some hours to fix `gz_rendering` build warnings about missing dependency `OptiX` only to find out in https://github.com/gazebosim/gz-rendering/issues/1010 that basically that renderer is ancient/broken... nevertheless, I think it would be good to leave here the bit and bolts that I discovered to be issue in my setup for `OptiX` discovery.

First of all, on Ubuntu 22.04 the library name is `libnvoptix.so.1` while this script currently searches for `liboptix.so` or `liboptix.1.so`, so the lookup would always fail IMHO.

Secondly, but it can be my fault, I tried setting:
- `--cmake-args OPTIX_INSTALL_DIR=path`
- `--cmake-args OptiX_INSTALL_DIR=path`
- export `OPTIX_INSTALL_DIR=path`

but the include files were never detected. I solved the hard way by directly checking the `$env` var in `find_path` call.

### DIFF
--- a/cmake/FindOptiX.cmake
+++ b/cmake/FindOptiX.cmake
@@ -83,12 +83,12 @@ endif()
 # gz-cmake modification: added "gz_" prefix to macro name
 macro(gz_OPTIX_find_api_library name version)
   find_library(${name}_LIBRARY
-    NAMES ${name}.${version} ${name}
+    NAMES nv${name}.${version} nv${name}
     PATHS "${OptiX_INSTALL_DIR}/lib${bit_dest}"
     NO_DEFAULT_PATH
     )
   find_library(${name}_LIBRARY
-    NAMES ${name}.${version} ${name}
+    NAMES nv${name}.${version} nv${name}
     )
   if(WIN32)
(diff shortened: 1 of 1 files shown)


j-rivero:
Thanks for the PR ! Are you interested into having the support in place even the rendering code can be broken with current versions?

Looking into the code:
```
set(OptiX_INSTALL_DIR "$ENV{OPTIX_INSTALL_DIR}" CACHE PATH "Path to OptiX installed location.")
```

Should be getting th OPTIX_INSTALL_DIR enviroment variable and put that in the correct place. In the worst case, I'm fine if we change all the ocurrences from `OptiX_INSTALL_DIR` to `$ENV{OPTIX_INSTALL_DIR}` but please be sure to delete the `CMakeCache.txt` and run the thing to check if the existing code is really working.

j-rivero:
Optix support was removed in https://github.com/gazebosim/gz-rendering/pull/1204. Close this one.

PR #497, opened 2025-04-28
GzCodeCoverage: ignore lcov errors

scpeters:
# 🦟 Bug fix

Fixes some errors related to code coverage on 24.04

## Summary

Running `make coverage` currently fails on 24.04 due to lcov errors and warnings. This is difficult to see because these warnings are currently redirected to `/dev/null`. Through debugging in https://github.com/gazebosim/gz-utils/pull/174, testing with nightly builds of gz-cmake5, and viewing GitHub workflow console logs ([for example](https://github.com/gazebosim/gz-utils/actions/runs/14674649013/job/41195440316)), several types of errors were identified that could be ignored to allow `make coverage` to pass for gz-utils.

There are other issues with code coverage, specifically with uploading coverage reports (similar to https://github.com/MicrosoftPremier/VstsExtensions/issues/237). I will open a separate issue about this.

## Checklist
- [ ] Added tests

### DIFF
--- a/cmake/GzCodeCoverage.cmake
+++ b/cmake/GzCodeCoverage.cmake
@@ -141,10 +141,14 @@ FUNCTION(gz_setup_target_for_coverage)
     # Capturing lcov counters and generating report
     COMMAND ${LCOV_PATH} ${_branch_flags} -q --no-checksum
       --directory ${PROJECT_BINARY_DIR} --capture
-      --output-file ${_outputname}.info 2>/dev/null
+      --ignore-errors gcov,gcov
+      --ignore-errors mismatch,mismatch
+      --ignore-errors unused,unused
+      --output-file ${_outputname}.info
     # Remove negative counts
     COMMAND sed -i '/,-/d' ${_outputname}.info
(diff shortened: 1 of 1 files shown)


cmake/GzCodeCoverage.cmake
```diff
@@ -141,10 +141,14 @@ FUNCTION(gz_setup_target_for_coverage)
     # Capturing lcov counters and generating report
     COMMAND ${LCOV_PATH} ${_branch_flags} -q --no-checksum
       --directory ${PROJECT_BINARY_DIR} --capture
-      --output-file ${_outputname}.info 2>/dev/null
+      --ignore-errors gcov,gcov
```
j-rivero, line 144: Does it make sense instead of ignoring categories doing it per files / per directories, something like:
 ```
  --exclude '/usr/include/*' \
  --exclude '*_TEST.cc' \
```

scpeters: we use `--remove` with lots of path wildcards on line 152. I'm not sure what the best approach is here. I think we would need to test with more packages as well so see if more `--ignore-errors` categories are needed

PR #589, opened 2026-09-20
Replace deprecated sre_compile with re

C88-YQ:
# 🦟 Bug fix

## Summary
[Python 3.11 deprecated](https://docs.python.org/3.12/whatsnew/3.11.html#:~:text=Undocumented%20modules%20sre_compile%2C%20sre_constants%20and%20sre_parse%20are%20now%20deprecated.%20(Contributed%20by%20Serhiy%20Storchaka%20in%20bpo%2D47152.)) the undocumented `sre_compile` module, causing a warning when running `make codecheck`:
```
DeprecationWarning: module 'sre_compile' is deprecated
```

This change replaces `sre_compile.compile` with the public `re.compile` API and removes the deprecated import.

## Backport Policy
- [ ] This is safe to backport to the following versions:
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [ ] Fortress
- [ ] This **should not** be backported
- [ ] I am not sure
- [ ] Other (fill in yourself)

## Checklist
- [ ] Added a screen capture or video to the PR description that demonstrates the fix (as needed)
- [ ] Added tests
- [x] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Assisted-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)


scpeters (commented):
it looks like there was a similar change in https://github.com/cpplint/cpplint/pull/214

can you look briefly at those changes and see if there are meaningful differences?

we should also consider importing a new version of cpplint from that repo

C88-YQ:
@scpeters Thanks a lot for the review and suggestions!

I took a look at cpplint/cpplint#214. It contains two commits:

1. replacing the deprecated `sre_compile` usage with `re` (same as the change in this PR)
2. removes cpplint's own regex cache and relies on Python `re`'s internal cache instead.

There was some discussion in that PR about the second change potentially causing a performance regression, so I did a small local test on `gz-sim`.

I ran `make codecheck` 10 times for both versions:

* Current version before dropping `sre_compile`: **24.3 s average**
* Version with both commits from cpplint/cpplint#214: **27.6 s average**

So there is some slowdown, although I think it is still within a reasonable range for `codecheck`.

For reference, I ran the test on the following system:

* CPU: Intel Core Ultra 5 125H (14 cores / 18 threads)
* RAM: 30 GiB
* OS: Ubuntu 24.04.4 LTS
* Kernel: 6.8.0-138-generic

C88-YQ:
Regarding importing a newer cpplint version, I agree that this would probably be a better long-term solution. I noticed that upstream cpplint is now at version 2.0.2, so this may be a good opportunity to update it.

I did a quick comparison and tried running the newer version locally. The changes are fairly large. Some are only formatting changes,  but there are also behavioral differences in the checks. I made a few small adjustments to the areas that I thought needed updating, but I still wasn't able to get `gz-sim` to produce results comparable to the current version. I think this would need a more careful and comprehensive review before updating it.

Given that, perhaps it would make more sense to close this PR and open a separate PR to update cpplint directly to 2.0.2.

@steve, what do you think would be the preferred direction here?

scpeters (approved):

scpeters:
> @scpeters Thanks a lot for the review and suggestions!
> 
> I took a look at [cpplint/cpplint#214](https://github.com/cpplint/cpplint/pull/214). It contains two commits:
> 
> 1. replacing the deprecated `sre_compile` usage with `re` (same as the change in this PR)
> 2. removes cpplint's own regex cache and relies on Python `re`'s internal cache instead.
> 
> There was some discussion in that PR about the second change potentially causing a performance regression, so I did a small local test on `gz-sim`.
> 
> I ran `make codecheck` 10 times for both versions:
> 
> * Current version before dropping `sre_compile`: **24.3 s average**
> * Version with both commits from [Drop deprecated sre_compile usage cpplint/cpplint#214](https://github.com/cpplint/cpplint/pull/214): **27.6 s average**
> 
> So there is some slowdown, although I think it is still within a reasonable range for `codecheck`.

yeah, this is a small change; I think it makes sense to just merge this and open a separate issue to evaluate updating the cpplint version

thanks for the contribution!

scpeters:
I'm going to let this live on main for a week before backporting

